Issue #162, opened 2014-11-21
Code cleanup checklist


jhamman:
Now that 4.2 has been released, we can move forward on cleaning up the source code. I spent a bit of time reading back through our discussion in #7, #48, and #134 and will try and provide an outline of what we need to do in order to close those issues.  One common thread that I think we want to stick to is to only keep documentation that is relevant and needed, extra stuff just leads to confusion down the road.  

**Code Cleanup:**
- [x] Remove unused variables (#134)
- [x] Remove relic "commented out" code (e.g. [this example](https://github.com/UW-Hydro/VIC/blob/develop/src/frozen_soil.c#L175-L193))
- [x] Remove hard coded literals (e.g. 905ef484b35454ea492cf9e20193b4a243b9d3e2)
- [x] Fix include statements (#163)

**Code Formatting (#48):**
- [x] Run [Uncrustify](http://uncrustify.sourceforge.net/) of all source files in repository.  @bartnijssen has already put together an [uncrustify options template](https://github.com/UW-Hydro/VIC/blob/master/tools/code_format/uncrustify_VIC_c.cfg).
- [x] Implement a consistent file header format.  In my opinion we should adopt a common format that would also play nice with code documentation tools like [`doxygen`](http://www.stack.nl/~dimitri/doxygen/index.html).  This would include the following pieces:
  - filename
  - original author(s)
  - description
  - license boilerplate
  
  **_Example header format:**_
  
  ``` C
  /*
  * \section DESCRIPTION
  *
  * A short description 
  *
  * \section LICENSE
  *
  * The Variable Infiltration Capacity (VIC) macroscale hydrological model
  * Copyright (C) 2014  The Land Surface Hydrology Group, Department of Civil
  * and Environmental Engineering, University of Washington.
  *
  * The VIC model is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
  * as published by the Free Software Foundation; either version 2
  * of the License, or (at your option) any later version.
  *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License along with
  * this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
  ```
[...]

Issue #36, opened 2013-08-07
Cumbersome input and output variable declarations in vicNl_def.h


tbohn:
Currently, input and output variable declarations are handled in vicNl_def.h via precompile defines that alias strings like "OUT_RUNOFF" to numbers.  The cumbersome part of this is that, if we wish to keep the list organized in some way for ease of finding variables (e.g., alphabetic and/or grouped by module), then adding a new variable in the middle of the list requires renumbering the entire remainder of the list.  This is tiresome and error-prone.  It also creates a lot of false differences between consecutive versions of vicNl_def.h, obscuring the true functional differences.

When I originally implemented this system, I did so only because my attempt to do this with an enumeration failed.  I'm not sure why; it should be possible (as in DHSVM).  At this point I think it is worthwhile to investigate the feasibility of switching to an enumeration.

PR #174, opened 2014-12-08
Adds CN to classic driver.  Still includes verbose option.

jhamman:
Initial pull request for adding CN to VIC.

What are our (@bartnijssen and @mabrunke) thoughts on: 
- Basic implementation,
- Portability,
- Documentation,
- Location of `CN/src` code, and 
- Merging recent cleanup changes to the `feature/CN` branch.

### DIFF
--- a/cn/src/CNAllocationMod.F90
+++ b/cn/src/CNAllocationMod.F90
@@ -1,0 +1,872 @@
+module CNAllocationMod
+
+!-----------------------------------------------------------------------
+!BOP
+!
+! !MODULE: CNAllocationMod
+!
+! !DESCRIPTION:
+! Module holding routines used in allocation model for coupled carbon
+! nitrogen code.
+!
+! !USES:
+  use shr_kind_mod, only: r8 => shr_kind_r8
+  implicit none
+  save
+  private
+! !PUBLIC MEMBER FUNCTIONS:
+  public :: CNAllocationInit    ! Initialization
+  public :: CNAllocation        ! run method
+
+! !PUBLIC DATA MEMBERS:
+   character(len=*), parameter, public :: suplnAll=& ! Supplemental Nitrogen for all PFT's
+                     'ALL'
+   character(len=*), parameter, public :: suplnCrp=& ! Supplemental Nitrogen for prognostic Crop
+                     'PROG_CROP_ONLY'
+   character(len=*), parameter, public :: suplnNon=& ! No supplemental Nitrogen
+                     'NONE'
+   character(len=15), public :: suplnitro = suplnNon ! Supplemental Nitrogen mode
+! !PRIVATE DATA MEMBERS:
+   real(r8):: bdnr                          !bulk denitrification rate (1/s)
+   real(r8):: dayscrecover                  !number of days to recover negative cpool
+   real(r8), pointer :: arepr(:)            !reproduction allocation coefficient
+   real(r8), pointer :: aroot(:)            !root allocation coefficient
+   real(r8), pointer:: col_plant_ndemand(:) !column-level plant N demand
+   logical :: Carbon_only = .false.         ! Carbon only mode 
+                                            ! (Nitrogen is prescribed NOT prognostic)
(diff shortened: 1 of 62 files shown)


jhamman:
Ok, I just finished a first read through the code and here are some initial comments:

##### Logistics:
- Update branch using current develop branch.  `feature/CN` branch is now quite a ways out of date with the`develop` branch.  The update/merge will require a bit of work but will make comparing to the `develop` or `driver/RASM` branches much easier.
- After the merge, the following tasks will need to be completed:
    - [ ] Convert "`#define`" variable names to `enum`s where applicable (see #36 and #175).  This applies to "Carbon model choices" and "Output Variable Types".
    - [ ] Use logging functions rather than print statements.
    - [ ] Add documentation to function and structure members (see example in #162)

##### Code Review/Style/Comments:
- We should think about how the C/Fortran interface looks.  `vic2clmtype_()` takes 192 arguments which screams fragile.  @bartnijssen and Tony Craig may have suggestions on how to do this.
- Handle error codes.  For example, [here in `vic_run()`](https://github.com/mabrunke/VIC/blob/feature/CN/vic_run/src/vic_run.c#L370-L374),
    ```C
    if(options.CARBON == CN_NORMAL || options.CARBON == CN_ADECOMP) {
        ErrorFlag = VICCNInterface(rec, Nveg, npfts, dmy, gp, atmos, all_vars, soil_con, veg_con, veg_lib, cn);
    ...
``` 
  Once the branch is updated to the current develop branch, `log_err()` will be available allowing:
   ```C
if (!ErrorFlag) {
        log_err("Error calling VICCNInterface");
}
```

- Variable types and memory usage.   `cn_data_struct` is quite large  and includes only `double` precision variables.  
  - What is `21` (e.g. onset_flag[21])?  It would probably be better use a global size variable `#define MAX_CN_LAYERS 21` and `onset_flag[MAX_CN_LAYERS]`
  - Use integer or bool datatypes for flags and counters (e.g. `double onset_flag[21];` --> `bool onset_flag[21];`
- Move "Record Carbon Cycling Variables" out of "Collect Water Balance Terms" function in `put_data.c`.  Rather than putting this code in an existing function, its probably worth creating a `collect_carbon_terms()` function that is only called when `CARBON == TRUE`.
- Remove hardcoded paths from Makefile
- Remove verbose comments
- Remove hard coded literals when possible from `cn/src/VICCNInterface.c`.  Physical constants are in `vic_run/include/physical_constants.h`

##### Relevant Upcoming Changes:
- Timestep units in `global_param.dt` are changing from hours to seconds.
- MPI in `driver/image`.  @bartnijssen can comment on which, if any changes will be necessary in this regard.

##### Next Steps:
Once we've settled on the changes to the classic driver, the next step is to port to the image mode and RASM drivers.

tbohn:
I agree with moving "collect carbon terms" into its own function - I should
have done it that way to begin with!

bartnijssen:
Some more comments with respect to organization (I haven't gone through the actual code yet):

 * Cleanly separate the CLM code that will be imported as a "block" or module (i.e. we will make no changes inside this) from the wrapper around it that will let this module be called from VIC. 

 * Separate all the I/O from the physics (akin to `vic_run()` and the different drivers). We should be able to use the `CN` code with all the drivers (even if we originally implement it for only a subset), without any changes to the physics code.

 * Determine where all the pieces will live within the vic repo directory structure. In VIC5 this will have to be broken out into separate pieces rather than as 60+ files in `cn/src`

jhamman:
@mabrunke - I'm going to close this PR.  Can you address the issues that @bartnijssen and I have brought up and open a new PR against this branch: https://github.com/UW-Hydro/VIC/tree/driver/cesm. 

Also, while you are moving things around a bit, we have reorganized the VIC directory structure a bit to better support adding extensions into VIC, such as CN. This PR is a good example of how we're structuring things: https://github.com/UW-Hydro/VIC/pull/231.  Following how the extension was added in #231, you'll want to add two extensions:
- dynveg_stub
- synveg_cn

PR #202, opened 2015-01-22
Hotfix/4.2.2

### LINKED ISSUE #201: Model Crashes When OUTPUT_FORCE=TRUE
When using `OUTPUT_FORCE = TRUE`, the model crashes with a `segmentation fault: bad access`.

Looking into this a bit deeper, this problem is coming in two locations where access to the `veg_con` structure, isn't protected by a `if (!options.OUTPUT_FORCE) {}` block.  This occurs in two blocks of the `initialize_atmos()` code:

1.  [Assign N_ELEM for veg-dependent forcings](https://github.com/UW-Hydro/VIC/blob/master/src/initialize_atmos.c#L257-L260),
2.  [Albedo, LAI, and Fractional Veg Cover](https://github.com/UW-Hydro/VIC/blob/master/src/initialize_atmos.c#L1424-L1589).

I've already applied a fix locally but this is probably going to mean we release a `hotfix/4.2.b` tag.

jhamman:
fixes for OUTPUT_FORCE=TRUE.

closes #201

### DIFF
--- a/src/ChangeLog
+++ b/src/ChangeLog
@@ -26,6 +26,28 @@ Usage:
 
 
 -------------------------------------------------------------------------------
+***** Description of changes between VIC 4.2.1 and VIC 4.2.2 *****
+-------------------------------------------------------------------------------
+
+Bug Fixes:
+----------
+
+Fixed memory error in initialize atmos when OUTPUT_FORCE = TRUE.
+
+	Files Affected:
+
+	initialize_atmos.
+	Makefile
+
+	Description:
+
+	Previously, access to unitialized elements of the veg_con and veg_hist
+	structure was attempted when OUTPUT_FORCE = TRUE, causing a memory error
+	and the model to crash.  This fix sets these elements inside a
+	`if (!options.OUTPUT_FORCE)` block allowing the OUTPUT_FORCE option to work as expected.
+
+
+-------------------------------------------------------------------------------
 ***** Description of changes between VIC 4.2.0 and VIC 4.2.1 *****
 -------------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


src/ChangeLog
```diff
@@ -26,6 +26,28 @@ Usage:
...
+
+	initialize_atmos.
+	Makefile
+
+	Description:
+
+	Previously, access to unitialized elements of the veg_con and veg_hist
+	structure was attempted when OUTPUT_FORCE = TRUE, causing a memory error
+	and the model to crash.  This fix sets these elements inside a
+	`if (!options.OUTPUT_FORCE)` block allowing the OUTPUT_FORCE option to work as expected.
```
bartnijssen: Break line so that it fits the general line width settings in the file

src/Makefile
```diff
@@ -154,7 +145,6 @@ full:
 	make depend
 	make tags
 	make model
-	make disagg
```
bartnijssen: This removed `vicDisagg` entirely as a compile-time option. Is that intentional? I assume that this can now be specified at run-time instead?

src/global.h
```diff
@@ -22,7 +22,7 @@
   2013-Dec-27 Removed QUICK_FS option.					TJB
   2014-May-20 Added ref_veg_vegcover.					TJB
 **********************************************************************/
-char *version = "4.2.1 2014-December-21";
+char *version = "4.2.2 2015-January-23";
```
bartnijssen: See comment about version numbering. I think the bug fix numbering scheme is inconsistent with 4.X (even if that is what we want for 5.X)

src/Makefile
jhamman: It was removed when @tbohn removed the `user_def.h` header file.  It is now specified in the global parameter file as `OUTPUT_FORCE = TRUE`.

jhamman:
this is ready now.

bartnijssen:
OK - just waiting for the travis build to finish

jhamman:
Thanks, I'll put the tag and email together shortly.

PR #212, opened 2015-05-06
Fixed indexing for veg_hist averages.

bartnijssen:
Corrected indexing (vidx) not (v) and moved within check for vidx != -1


bartnijssen:
Note that following this pull request, the develop branch will run through the Stehekin test case (albeit it with many warnings). This does not mean that the output make sense, but the model does not crash or abort for the Stehekin case (tested for 2 years).

PR #251, opened 2015-08-12
Feature/fix vic time

jhamman:


tests/unit/shared/test_vic_time.py
```diff
@@ -75,100 +85,153 @@ def test_fractional_day_from_dmy(dmy_feb_3_noon):
...
 @pytest.mark.skipif(nctime_unavailable, reason=nc_reason)
 def test_make_lastday():
-    lastday = np.zeros(12)
+    # wasn't able to map lastday to a numpy array, don't know why...
+    lastday = ffi.new('unsigned short int []', [0] * 12)
     for year in np.arange(1900, 2100):
-        for cal in ['standard', 'gregorian', 'proleptic_gregorian', 'julian']:
-            vic_lib.make_lastday(calendars[cal], year,
-                                 ffi.addressof(lastday).ctypes.data)
+        for cal in ['standard', 'gregorian', 'proleptic_gregorian']:
```
bartnijssen: Did you drop `julian ` for a reason?

jhamman: yes, the julian calendar doesn't have a leap year in 1990 and I didn't have a programmatic way to test against that.

bartnijssen: But isn't that the one that you would need to test the date2num functionality in VIC or is that a separate issue. If we cannot test for it, should we disallow the calendar in VIC? Or just accept the PR and open an issue to fix the test in the future?

jhamman: This is just testing whether an individual year is a leap year or not.  For the standard calendars (`['standard', 'gregorian', 'proleptic_gregorian']`), we can use the python calendar module to test against.  The julian calendar isn't available to us via the `calendar` package.  

I think the date2num and num2date tests have sufficiently shown that we are getting the same answers as the `netcdftime` package.

PR #253, opened 2015-08-12
Minor edits to documentation

bartnijssen:
I haven't gone through all of it - mostly through the higher level stuff (about vic) rather than the instructions for using VIC. Mostly I toned down the UW role, since we want this to be a community effort and removed things that are likely to date quickly.

However, I think this is sufficient for a start. At a minimum it will duplicate the functionality of the existing UW VIC page on read the docs and we gain community access (and hopefully involvement).


bartnijssen:
@jhamman: Please go ahead and review - if it is ready to be merged than go ahead and merge the original PR as well

jhamman:
Just looking through it now.  Can you remove `docs/Development/ModelDevelopment.md` from `mkdocs.yml`?

After that, I'll merge.  I'm going to add a readme to the docs dir that explains how to edit and test the docs in the future.

PR #265, opened 2015-08-31
CESM driver

### LINKED ISSUE #196: CESM Driver Todo List
This will be a running to do list as we work on coupling VIC.5 in CESM.  @bartnijssen, @apcraig, and @mabrunke - feel free to make additions or comments to this list.  

#### Infrastructure to Port from VIC-RASM.1.0
- [ ] Monthly averaging: `wrf_write_data.c`, if only in concept (#198, #281)

#### Science/Physics Options and Features to Port from VIC-RASM.1.0
- [ ] Vegetation dependent snow albedo (RACM: r370)
- [ ] Longwave/emissivity (RASM: r790)
- [ ] Lowest WRF model level height (RASM: r?)

#### New Infrastructure:
- [x] CESM build system integration
- [x] C-Fortran Interface
  - [x] `vic_cesm_init_mpi(MPI_COMMUNICATOR);`
  - [x] `vic_cesm_init(clock, case_metadata);` :  Initialization call from `lnd_init_mct()`.
  - [x] `vic_cesm_run(clock);`:  Run call from `lnd_run_mct()` including forcing data from the coupler `x2l` and `l2x` returning fluxes.
  - [x] `vic_cesm_final();`:  Free memory, this can be called directly from Fortran.

#### New Science/Physics Features
- [ ] Carbon and Dynamic Vegetation (CN-DV) from @mabrunke. (#174)

#### Testing (in addition to other VIC testing in #79 and #177)
- [ ] Exact restarts in CESM
- [ ] Image mode driver results match `RI` compset results

#### Last Steps
- [ ] Initial state generation for VIC and eventually, CN

jhamman:
closes #196


vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
```diff
@@ -0,0 +1,722 @@
...
+    !--- set shr unit number to vic log file
+    CALL shr_file_getLogUnit(shrlogunit)
+    IF (mpicom_lnd == 0) THEN
+       ! Tony look here INQUIRE(file='lnd_modelio.nml'//TRIM(inst_suffix), exist=exists)
+       exists = .FALSE.
+       IF (exists) THEN
+          iulog = shr_file_getUnit()
+          !Tony look here CALL shr_file_setIO('lnd_modelio.nml'//TRIM(inst_suffix), iulog)
+          CALL shr_file_setIO('lnd_modelio.nml', iulog)
+       ENDIF
```
jhamman: @apcraig - We are creating per task log files inside VIC.  Do we need this section?  Should we have a separate log file for the driver?  I think it is pretty difficult to share file handles between C and Fortran.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
```diff
@@ -0,0 +1,722 @@
...
+    vclock%current_month = month
+    vclock%current_day = day
+    vclock%current_dayseconds = seconds
+    vclock%state_flag = seq_timemgr_RestartAlarmIsOn(EClock)
+    vclock%stop_flag = seq_timemgr_StopAlarmIsOn(EClock)
+
+    !--- get time of next radiation calc for time dependent albedo calc
+    ! Tony - is this for the coupler or for the land model.  We dont need it
+    ! in VIC
+    !CALL seq_infodata_GetData(infodata, nextsw_cday=nextsw_cday)
```
jhamman: @apcraig - Can you comment on the need for this line?

jhamman:
@apcraig - The Fortran source files in this PR are ready for a review.  I drew attention to a few outstanding questions but a full review of `lnd_comp_mct.F90` would be great.

jhamman:
Here is a summary of the functionality of the current code:
- Model builds in RI compset
- Initializes memory and mpi in C and Fortran
- Passes fluxes between coupler and VIC.  VIC to coupler fluxes are just missing values at this point but the infrastructure is in place in `lnd_comp_mct`
- time is advanced in VIC and is checked against coupler values each timestep.  This is all done inside VIC. An instructive error message is raised and an `exit(1)` is called if the clocks are out of sync.
- state and stop flags are passed

Big things left to do:
- Build setup infrastructure for configuration files (Maybe I could @apcraig to help with this).
- Fill `l2x` fields inside VIC
- Reading/writing state files

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
apcraig: The driver and other components manage their own log files.  In some ways, vic needs to decide how to handle this.  In general, in CESM, a component should write from the root processor to the *.log file associated with lnd_modelio.nml and other tasks in that component can write to unit number 6 (which is the ccsm.log file usually).  If VIC does not want to follow that convention, that's fine.  In this case, I propose the vic driver (lnd_comp_mct) and other vic coupling layer methods use this approach and that vic "internal" write to it's own files.   By the way, in this code section, the logic should not be "IF (mpicom_lnd == 0) THEN", it should be something like
iulog = 6
if (my_task == 0) then
          iulog = shr_file_getUnit()
          CALL shr_file_setIO('lnd_modelio.nml', iulog)
endif

In other words, all tasks will write to unit 6 through the iulog unit number except the root task.  The root task will "get" a unique unit number and then assign that unit number of the lnd.log file.

One other thing that we could think about is managing the unit numbers in vic.  If there are any, we could query for free unit numbers in the init call in lnd_comp_mct and assign them to vic internal variables.  But this might not be necessary with C, not sure.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
apcraig: Just leave it there and leave it commented out.  This is important IF vic has time dependent (solar angle dependent) albedos.  If it does, then we need to pass this value into vic and it needs to use it.  Basically, nextsw_cday is the julian day of the next atmosphere shortwave radiation calculation.  That "time" should be used to compute the albedos.  In other words, the atm sets this value and passes it to surface components.  Those surface components are supposed to use this time to "place the sun in the sky" for their albedo computations to pass back to the atm model for the radiation calculation.  If there is no time (solar angle) dependence for surface albedos in VIC, nextsw_cday can be ignored.  But I recommend leaving it as is, commented out, as a reminder in case it's needed in the future.

apcraig:
I'm looking at the lnd_comp_mct file on the driver/cesm branch.  Everything looks pretty good.  I have some small changes for iulog and shrlogunit unit number management.   I also will update some of the subname and write statements for clarity and consistency.  I have created a fork, pushed to my fork, and submitted a pull request.

Where is the check for clock synchronization?  Is that inside VIC?  It looks like vclock is filled with driver values and that gets passed into the run method.  Normally, the clock check would be at the end of the run method.  By the way, the time in the driver is the time at the END of the coupling period.

vic/drivers/cesm/src/cesm_interface_c.c
```diff
@@ -0,0 +1,181 @@
...
+ *****************************************************************************/
+int
+vic_cesm_run(vic_clock vclock)
+{
+    // reset l2x fields
+    initialize_l2x_data();
+
+    // advance the clock
+    advance_time();
+    assert_time_insync(&vclock, &dmy);
```
jhamman: @apcraig - here is the clock check you were asking about. I think this is right, given that `vclock` is representing the time at the end of the coupling period.

jhamman: We could do this check in `lnd_comp_mct.F90` if that is better.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
jhamman: Sounds good.  I think we're on the same page here.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
jhamman: It seems passing streams or open file descriptors between C and Fortran is a non starter :
- http://stackoverflow.com/questions/18466823/c-c-and-fortran-writing-to-same-file
- http://computer-programming-forum.com/49-fortran/c14c1198dd713790.htm

So I think we're going to have to follow this plan: 
- The `my_task == 0` process in `lnd_comp_mct` writes to the `my_task == 0` defined by `lnd_modelio.nml` 
- All other processes in `lnd_comp_mct` write to unit number 6 (or whatever the shared log unit is)
- Each VIC process, including `my_task == 0` in C writes to its own log file

vic/drivers/cesm/src/cesm_interface_c.c
apcraig: It's fine on the c side.  As long as it's clear that it's happening and is working correctly.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
apcraig: That sounds like a sound plan.  thanks,

tony.....

vic/drivers/cesm/src/init_library.c
```diff
@@ -0,0 +1,251 @@
...
+        l2x_vic[i].l2x_Fall_swnet = MISSING;
+        l2x_vic[i].l2x_Fall_fco2_lnd = MISSING;
+        l2x_vic[i].l2x_Fall_flxdst1 = MISSING;
+        l2x_vic[i].l2x_Fall_flxdst2 = MISSING;
+        l2x_vic[i].l2x_Fall_flxdst3 = MISSING;
+        l2x_vic[i].l2x_Fall_flxdst4 = MISSING;
+        l2x_vic[i].l2x_Fall_flxvoc = MISSING;
+        l2x_vic[i].l2x_Flrl_rofliq = MISSING;
+        l2x_vic[i].l2x_Flrl_rofice = MISSING;
+        l2x_vic[i].l2x_vars_set = false;
```
jhamman: @apcraig - Do you think it is better to return a missing value or zero for `l2x` fields that we are not using?  Is there a convention for this?

apcraig: Generally we set unused fields and masked regions in a field to a special value.  That value is defined in shr_const_mod and is called shr_const_spval.  It happens to be 1.0e30 right now.  It is not a requirement, but it is recommended because it makes trapping problems more likely.  I think what you're doing is fine.  I'm not sure what value MISSING is taking on.  It probably should not be a nan or infinity or undefined because we allow masked operations on those values and we don't want to trap those operations.  So MISSING should probably be some large positive or negative value as suggested by shr_const_spval.

vic/drivers/cesm/bld/vic.buildexe.csh
```diff
@@ -0,0 +1,23 @@
...
+$CODEROOT/lnd/vic/vic/vic_run/src
+$CODEROOT/lnd/vic/vic/vic_run/include
+$CODEROOT/lnd/vic/vic/drivers/shared/src
+$CODEROOT/lnd/vic/vic/drivers/shared/include
+$CODEROOT/lnd/vic/vic/drivers/cesm/src
+$CODEROOT/lnd/vic/vic/drivers/cesm/include
+$CODEROOT/lnd/vic/vic/drivers/cesm/cpl_$comp/
+EOF
+
+set vicdefs = ""
```
jhamman: @apcraig - 

What is the best way to set build time flags and options using the CESM Makefile?  Is this the line to do it in?

We are currently setting some basic C compiler flags: `CFLAGS  =  -ggdb -O0 -Wall -Wextra -fPIC`

and a build time log level flag: `-DLOG_LVL=$(LOG_LVL)` where `$(LOG_LVL)` is an integer that sets the verbosity of logging within VIC.

vic/drivers/cesm/bld/vic.buildnml.csh
```diff
@@ -0,0 +1,3 @@
+#! /bin/csh -f
+
+cd $RUNDIR
```
jhamman: @apcraig - 

I think this is going to be my next task.  Can you walk me through how you think we should configure the namelist build step?  VIC has one "namelist" that we call a global configuration file.  It is just a simple ASCII file that looks something like this:

```
# VIC Model Parameters for CESM driver

# Simulation Parameters
SNOW_STEPS_PER_DAY	72
RUNOFF_STEPS_PER_DAY  72

FULL_ENERGY TRUE
FROZEN_SOIL	TRUE

# Simulation Options
AERO_RESIST_CANSNOW AR_406

# DOMAIN INFO
DOMAIN ${DIN_LOC_ROOT}lnd/vic/griddata/fracdata_wr50a_ar9v4_101013.nc

FULL_ENERGY TRUE
FROZEN_SOIL	FALSE
```

So we'll have a default global config file that needs to have the variables like `${DIN_LOC_ROOT}l` expanded. I can take care of all that in `vic.buildnml.csh`.  There is also `user_nl_vic` in the `${CASE}` directory.  What is the CESM standard on how to use that? Ideally, we won't need to edit `vic.buildnml.csh` to change the VIC configuration, right?

apcraig: You can probably just "cat" the file and then make sure it ends up in the $RUNDIR.  We might need to make some other changes so it also gets copied to the CaseDocs file.  That happens in preview_namelists.  You also should try to generate a vic.input_data.list file containing the filenames the model requires for this run.  That scripts check that all those files exist before trying to run as a way to exit gracefully on missing data.  You do not need to have ongoing restart file there.  In general, you should be able to get the coupling period from $LND_NCPL.  Have a look at the current vic.buildnml.csh for all these things, you should be able to replicate that example for your input file (vs a namelist file).  user_nl_vic is not useful.  We should make sure that file is not generated in the case directory.

vic/drivers/cesm/bld/vic.buildexe.csh
apcraig: You are going to have to push that change into the Macros file and the options will probably depend on the compiler.  You will modify scripts/ccsm_utils/Machines/config_compilers.xml.  Have a look there and see if you can figure it out.  I don't think there is much documentation unfortunately.  there might need to be some trial and error.  In the end, you want something like 
  <CFLAGS MODEL="vic"> -ggdb -O0 -Wall -Wextra -fPIC </ADD_CFLAGS>
somewhere and you want the Macros file to end up with something like
ifeq ($(MODEL), vic) 
   CFLAGS += --ggdb -O0 -Wall -Wextra -fPIC
endif

vic/drivers/cesm/bld/vic.buildnml.csh
jhamman: @apcraig - Thanks for the tips.  I ended up getting a bit fancier than I originally intended but I like what we have now.  Default namelist files are copied from the `vic/drivers/cesm/bld` directory to `$CASE/BuildConf/vicconf` where they are editable on a case by case basis.  They are also copied, in a stripped down format to the `$RUNDIR`.  I am writing a `vic.input_data_list` file and that seems to be working right because other parts of the setup scripts are complaining that I haven't moved all the inputs to Spirit yet.  

My only remaining question is, what do we want in `$CASE/CaseDocs`?  Read only copies of VIC namelists (there are two, sort of...)?  It is quite easy to do that.

vic/drivers/cesm/bld/vic.buildexe.csh
jhamman: Okay, I'm going to leave this for now and may ask you to step in if we end up having problems with the build without these flags.  Currently, the only flags we're getting with `icc` are: `icc -c -I... -O1 -fp-model precise   -DLINUX  -DMCT_INTERFACE -DHAVE_MPI -DFORTRANUNDERSCORE -DNO_R16 -DHAVE_F2008_CONTIGUOUS -DLINUX -DCPRINTEL  -DHAVE_SLASHPROC`.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
```diff
@@ -0,0 +1,729 @@
...
+    INTEGER(C_INT)                              :: errno
+    CHARACTER(len=VICMAXSTRING, kind=C_CHAR)    :: vic_global_param_file
+
+    !--- initialize the field index values, subroutine is below in this file
+    CALL cpl_indices_set()
+
+    !--- point/get data from cdata datatype
+    CALL seq_cdata_setptrs(cdata, ID=LNDID, mpicom=mpicom_lnd, &
+         gsMap=GSMap_lnd, dom=dom_lnd, infodata=infodata)
+    CALL MPI_COMM_RANK(mpicom_lnd, mytask, ierr)
```
jhamman: @apcraig - I'm debugging this code now and ran across this issue: If I print `mpicom_lnd`, most processes show a value of 11, but two, show a value of 12.  I figured all the mpi communicators would be the same.  Am I missing something or is this an issue?

vic/drivers/cesm/bld/vic.buildexe.csh
jhamman: Turns out this was pretty simple, the change was just
```xml
<compiler COMPILER="intel">
  ...
  <ADD_CFLAGS MODEL="vic">  -std=c99 </ADD_CFLAGS>
  <ADD_CFLAGS MODEL="vic" DEBUG="TRUE"> -Wall -traceback -DLOG_LVL=0 </ADD_CFLAGS>
</compiler>
```
yielding a this in the Macros file:
```csh
ifeq ($(DEBUG), TRUE) 
  ifeq ($(MODEL), vic) 
     CFLAGS += -Wall -traceback -DLOG_LVL=0 
  endif

   CFLAGS += -g 
   FFLAGS += -g -CU -check pointers -check bounds -fpe0 -ftz 
endif

...

ifeq ($(MODEL), vic) 
   CFLAGS +=  -std=c99 
endif
```
pretty slick I'd say.

vic/vic_run/include/vic_log.h
```diff
@@ -54,7 +54,7 @@
 // | INFO      | < 20             |
 // | DEBUG     | < 10             |
 #ifndef LOG_LVL
-#define LOG_LVL 25
+#define LOG_LVL 5
```
jhamman: reminder to change back to 25 after debugging / changes to `ccsm_utils/Machines/config_compilers.xml` are committed.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
apcraig: An mpi communicator ID is not necessarily the same on all tasks.  That's a feature, and I agree that it's confusing.  Think of it as a pointer to some local data that is "common" across tasks.  The pointer ID can be different on different tasks, but the data it points to will be consistent with the data on other tasks.

vic/drivers/cesm/bld/vic.buildnml.csh
apcraig: CaseDocs just contains read-only copies of input files in the run directory.
It's a place to store copies, that's all.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
jhamman: Interesting and yes, a bit confusing.  At least its not a bug.  I've moved on and everything seems to be okay at the MPI level. (We've got plenty of other problems!)

vic/drivers/cesm/bld/build_vic_namelist
```diff
@@ -0,0 +1,164 @@
...
+    # Get the config for the current LND_GRID
+    try:
+        grid_config = all_grids[grid]
+    except KeyError:
+        raise KeyError('Requested Grid "{0}" is not '
+                       'defined in {1}'.format(grid, config_file))
+
+    # copy VIC files to confdir
+    for key in ['vic_global_param', 'vic_constants']:
+        grid_config[key] = copy_to_vicconf(coderoot, confdir, grid_config[key])
```
jhamman: TODO: copy read only versions to CaseDocs dir.

vic/drivers/cesm/cpl_mct/lnd_comp_mct.F90
```diff
@@ -0,0 +1,729 @@
...
+    TYPE(location_struct), DIMENSION(:), POINTER :: locations
+    CHARACTER(len=*), PARAMETER :: subname = '(unpack_vic_domain)'
+
+    !--- Associate locations pointer in local_domain structure
+    CALL c_f_pointer(local_domain%locations, locations, [local_domain%ncells])
+
+    !--- Get domain information for each cell in local domain
+    DO i = 1, local_domain%ncells
+       gindex(i) = f_index(locations(i)%global_idx, &
+                           global_domain%n_nx, global_domain%n_ny)
```
jhamman: @bartnijssen - We need to figure out how to scatter the `global_domain` structure or at least the global domain shape (`n_nx` and `n_ny`).  We need the global domain in two places at the Fortran level:
- Here, when we translate the 1d index from C order to Fortran order.
- Line 247 (above) when we calculate the full domain size / shape.

I got started on creating a MPI type for the `domain_struct` but got hung up on the `locations` member.  It seems difficult to create an MPI type of a structure with an allocatable array (pointer) inside. I understand why, I'm curious though if you have a preferred way to proceed in this area.  

The good news, after duct taping this section of code together, we're through the lnd_init step.  Not as good news is we are crashing in at the driver cesm driver level now. More work to be done.

refs:
- http://stackoverflow.com/questions/10419990/creating-an-mpi-datatype-for-a-structure-containing-pointers
- http://stackoverflow.com/questions/19637159/creating-structs-and-unions-on-mpi/19641945#19641945

vic/drivers/cesm/bld/build_vic_namelist
jhamman: done in 628f7b2

jhamman:
This last commit gets us into `vic_run`, currently failing in `CalcAerodynamic.c`:

```
forrtl: error (73): floating divide by zero
Image              PC                Routine            Line        Source             
cesm.exe           000000000053A818  CalcAerodynamic            78  CalcAerodynamic.c
cesm.exe           000000000052E138  vic_run                   300  vic_run.c
cesm.exe           0000000000524EF9  vic_cesm_run_mode          54  vic_cesm_run.c
cesm.exe           000000000051F426  vic_cesm_run              112  cesm_interface_c.c
cesm.exe           000000000050B80A  lnd_comp_mct_mp_l         393  lnd_comp_mct.F90
cesm.exe           000000000040352D  ccsm_comp_mod_mp_        2900  ccsm_comp_mod.F90
cesm.exe           0000000000453BB3  MAIN__                     91  ccsm_driver.F90
```

jhamman:
The travis builds are working again here and the failing tests indicate bugs in this branch.

vic/drivers/cesm/src/cesm_interface_f.F90
```diff
@@ -0,0 +1,99 @@
...
+  END INTERFACE
+
+  !--------------------------------------------------------------------------
+  !> @brief   Init MPI Interface
+  !--------------------------------------------------------------------------
+  INTERFACE
+     SUBROUTINE initialize_mpi(MPI_COMM_VIC_F) BIND(C, name='initialize_mpi')
+       USE, INTRINSIC :: ISO_C_BINDING
+       IMPLICIT NONE
+       INTEGER, INTENT(inout) :: MPI_COMM_VIC_F
```
jhamman: One remaining issue (warning) with the Fortran/C interoperability: 

```Fortran
./src/cesm_interface_f.F90:56.45:

     SUBROUTINE initialize_mpi(MPI_COMM_VIC_F) BIND(C, name='initialize_mpi')

                                             1

Warning: Variable 'mpi_comm_vic_f' at (1) is a parameter to the BIND(C) procedure 'initialize_mpi' but may not be C interoperable
```

It may be possible to cast `MPI_COMM_VIC_F` to a `C_INT` to suppress this warning.

jhamman:
Getting close:

Failing in intialization: 

```
[ERROR] /work1/home/jhamman/rasm_vic5/models/lnd/vic/vic/drivers/cesm/src/vic_init.c:590: errno: None: Calculated wilting point moisture (0.000000 mm) is less than calculated residual moisture (99385147393784118539060484416789166515295969199253566394340348822709862400000.000000 mm) for layer 0.
        In the soil parameter file, Wpwp_FRACT MUST be >= resid_moist / (1.0 - bulk_density/soil_density).
location:
        latitude       : 16.5350
        longitude      : -170.7771
        area           : 0.0000
        frac           : 0.0000

        veg            : 0
        global_idx     : 0
        io_idx         : 0
        local_idx      : 0
```

bartnijssen:
@jhamman - is this one still open after we merged #397?

jhamman:
closed via #397

PR #287, opened 2015-09-15
Broadcast global_domain

bartnijssen:
Haven't had a chance to compile or run. @jhamman : please review

Problem:
 * global `nx` and `ny` not available on any process other than the master process

Solution:
 * define `MPI_Datatype` for `domain_struct`
 * broadcast `global_domain` (this is already available as a global on all the processes)
 * set `global_domain.locations` to `NULL` for anything other than the master process


jhamman:
Thanks, I fixed the build and merged into `driver/cesm` manually.

Issue #211, opened 2015-05-06
Specify netCDF variable names in global config file


jhamman:
In the current image mode driver, we have hard coded the variable names for forcing variables, domain variables, and parameter variables.  This includes netCDF dimension names.  As I have just moved to a new test case, I've had to modify these hard coded variable names, for example:
```diff
-    global_domain->n_nx = get_nc_dimension(nc_name, "ni");
-    global_domain->n_ny = get_nc_dimension(nc_name, "nj");
+    global_domain->n_nx = get_nc_dimension(nc_name, "lon");
+    global_domain->n_ny = get_nc_dimension(nc_name, "lat");
-    get_nc_field_int(nc_name, "run_cell", d2start, d2count, run);
+    get_nc_field_int(nc_name, "mask", d2start, d2count, run);
-    get_nc_field_double(nc_name, "xc",
+    get_nc_field_double(nc_name, "lon",
-    get_nc_field_double(nc_name, "yc",
+    get_nc_field_double(nc_name, "lat",
```

Would it make sense to add these field names to the global parameter file?  This is probably most important for the dimension names and for the forcing variables.  Parameter variables and domain file variables are mostly fixed and I don't think we need to provide too much flexibility in that way.  

@bartnijssen - The changes in the above diff were required to run the Stehekin example after the commits in #210.

bartnijssen:
They should all be configurable (including dimension names). That is, they should be read from a config file as a dictionary.

jhamman:
cc @wietsefranssen

jhamman:
my proposal for how we should do this in the global parameter file is that we specify the netCDF variable name after listing the variable identifier e.g. `FORCE_TYPE    <var_identifier>    <nc_var_name>`.  In practice, this would look like this:

```
FORCING1      (put the forcing path/prefix here  # Forcing file path and prefix, ending in "_"
FORCE_TYPE    AIR_TEMP 	   tas    # Average air temperature, C
FORCE_TYPE    PREC 	       prcp   # Total precipitation (rain and snow), mm
FORCE_TYPE    PRESSURE 	   pres   # Atmospheric pressure, kPa
FORCE_TYPE    SHORTWAVE    dswrf  # Incoming shortwave, W/m2
FORCE_TYPE    LONGWAVE 	   dslwr  # Incoming longwave radiation, W/m2
FORCE_TYPE    QAIR 	       shum   # Vapor pressure, kPa
[...]

PR #296, opened 2015-09-28
switched 'xc' and 'yc' in get_global_domain.c

wietsefranssen:
xc and yc need to be switched. See also ncdump of Stehekin_image_test.domain.nc:

```
double xc(nj=4, ni=5);
:standard_name = "latitude";
:long_name = "latitude";
:units = "degrees_north";
:axis = "Y";

double yc(nj=4, ni=5);
  :standard_name = "longitude";
  :long_name = "longitude";
  :units = "degrees_east";
  :axis = "X";
```

### DIFF
--- a/vic/drivers/image/src/get_global_domain.c
+++ b/vic/drivers/image/src/get_global_domain.c
@@ -112,7 +112,7 @@ get_global_domain(char          *nc_name,
 
     // get longitude -
     // TBD: read var id from file
-    get_nc_field_double(nc_name, "xc",
+    get_nc_field_double(nc_name, "yc",
                         d2start, d2count, var);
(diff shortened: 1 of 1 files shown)


jhamman:
This seems to be more of an issue in the test dataset and not with the VIC model.  

Ideally, we will specify the variable and dimension names in the global parameter file so we don't have these hard coded strings (see #211 for more info on that issue).

wietsefranssen:
Thats fine with me. I switched xc and xy in the testfile 'Stehekin_image_test.domain.nc', to make it compatible with the routing example file: 'stehekin_parameters_01.rvic.prm.Stehekin.20150727.nc'.

Issue #297, opened 2015-09-30
Add a time variable in NetCDF output file


wietsefranssen:

wietsefranssen:
Together with the time variable we also want to include the lat and lon variables in the NetCDF output.
We want to do this as much ad possible following the cf-conventions.

We prefer have one dimensional lat/lon variables like the standard way:
see chapter 5.1: http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#coordinate-system
however chapter 5.2 is also possible but less used and also less supported by our analysis software packages.

What is your opinion about this?

jhamman:
Yes, we want to be able to include the x (lon) and y (lat) coordinates in the output file.  We do however, need to be able to support 1 and 2 dimensional coordinates.  So the answer I think is both.

wietsefranssen:
the standard one dimensional way is:
        double lon(lon) ;
                lon:long_name = "longitude" ;
                lon:units = "degrees_east" ;
                lon:standard_name = "longitude" ;
        double lat(lat) ;
                lat:long_name = "latitude" ;
                lat:units = "degrees_north" ;
                lat:standard_name = "latitude" ;

the 2-dim ways is: 
float xc(xc) ;
xc:axis = "X" ;
xc:long_name = "x-coordinate in Cartesian system" ;
xc:units = "m" ;
float yc(yc) ;
yc:axis = "Y" ;
yc:long_name = "y-coordinate in Cartesian system" ;
float lon(yc,xc) ;
lon:long_name = "longitude" ;
lon:units = "degrees_east" ;
float lat(yc,xc) ;
lat:long_name = "latitude" ;
lat:units = "degrees_north" ;

this means that if we want to follow the protocol we only can choice between one of the two options the same time (naming conflict). In order to solve this we have three options:
1. rename the lon and lat var names in something else
2. user can choose in vic param file (or somewhere else?) which one to use.
3. VIC detects if the grid is rectangular or not (in first case it outputs the 1D ways, else it write the 2D way)

wietsefranssen:
An other thing (FYI): we need the lat lons of the full domain (no NANs because of the mask).
Thats why I includes the following lines in get_global_domain.c

    // if MASTER_PROC
    global_domain->locations_grid = (location_grid_struct *)
                               malloc(global_domain->n_ny * global_domain->n_nx *
                                      sizeof(location_grid_struct));
    if (global_domain->locations_grid == NULL) {
[...]

Issue #197, opened 2015-01-18
Image Mode Driver: Time Handling


jhamman:
Time handling in the image mode driver currently requires that the forcing file begins at the same time as the simulation.  Furthermore, the history files from the image driver do not have a time variable.

I propose using the recent time/calendar improvements in #188 to better handle the time variable in this driver.

jhamman:
cc @wietsefranssen

Issue #240, opened 2015-08-06
Overhaul the subdaily disaggregation logic in initialize_atmos.c


jhamman:
When we changed the default timestep units to seconds (#188), we introduced a number of bugs into the subdail disaggregation of forcing variables in `initialize_atmos.c`. Looking at the code, there is a lot of repeated logic that could be abstracted to individual functions.  This will greatly reduce the likelihood of bugs coming from slightly different implementations with the same goal. 

I'll take point on this issue and will split the work into three steps:
- [ ] Create functions for the various forms of subdaily disaggregation supported in `initialize_atmos`.
- [ ] Replace repeated sections of code with these functions
- [ ] Test, test, test.

xref #42 and #188 

cc @tbohn: I may need a bit of guidance as I go through this.

bartnijssen:
Can you point to some examples? The MTCLIM was wrapped with as few changes to that code as possible and I would like to keep it that way. Any VIC code we can of course change.

tbohn:
I think Joe is referring to the logic in initialize_atmos.c for
transferring data from the MTCLIM data structures and the forcing data
structures into the atmos data structure.

Ted

tbohn:
OK, as per @jhamman 's suggestion, I'll take a look at getting this subroutinized.

jhamman:
Alright, I think I've fixed some of these issues shown below but will have to rerun this test when I get back to Seattle.  A summary of what where I see the subdaily disaggregation (check these boxes when we are satisfied that things are working):

- [ ] Precipitation: looks right, need to plot subdaily subset
- [ ] Air Temp: looks right, need to plot subdaily subset
- [ ] Shortwave: Units problem, needs work
- [ ] Longwave: Units and indexing problem, needs work
- [ ] Density: broken
- [ ] Pressure: broken
- [ ] VP: broken
- [ ] Wind: looks right, need to plot subdaily subset

![stehekin_forcings](https://cloud.githubusercontent.com/assets/2443309/9913024/72a457a8-5c5f-11e5-83e9-48a4f06c5c73.png)
*Figure: Comparison of input forcings for one grid cell (48.1875, -120.6875) in the Stehekin test dataset with the disaggregated hourly forcings for VIC 4.2.b and the develop branch.*

Figure current as of January 26, 2015.

tbohn:
[...]

PR #303, opened 2015-10-19
Outfile Dimensions

wietsefranssen:
We have added lat, lon and time variables in the NetCDF output files. One can choose between 1D and 2D lat/lon variables by setting COORD_DIMS_OUT in the global parameter file.

Adapted time dimension in NetCDF output:
Number of timesteps was based on MODEL_STEPS_PER_DAY in the global parameter file.
Now the number of timesteps is based on OUTPUT_STEPS_PER_DAY in the global parameter file.


vic/drivers/image/include/vic_driver_image.h
```diff
@@ -63,6 +70,7 @@ typedef struct {
     size_t n_nx; /**< size of x-index; */
     size_t n_ny; /**< size of y-index */
     location_struct *locations; /**< locations structs for local domain */
+    location_grid_struct *locations_grid; /**< locations structs for full unmasked latlon domain */
```
jhamman: I'm wondering if there is a better way to do this.  Perhaps we could create a `location_struct` for all grid cells, even if they are not being run.  We would just need to add an `active` (or something similar) member to the `location_struct`.  When we do the decomposition, we would only split up active grid cells so nothing would change below the `MASTER_PROC` level.

vic/drivers/image/include/vic_driver_image.h
```diff
@@ -187,6 +203,7 @@ void vic_start(void);
 void vic_store(void);
 void vic_write(void);
 char will_it_snow(double *t, double t_offset, double max_snow_temp,
-                  double *prcp, size_t n);
+                  double *prcp,
+                  size_t n);
```
jhamman: are all of these whitespace changes from uncrustify?  They don't seem necessary.  Were you using the uncrustify configuration file in the VIC repo?

vic/drivers/image/src/get_global_param.c
```diff
@@ -838,6 +841,9 @@ get_global_param(FILE *gp)
             else if (strcasecmp("N_OUTFILES", optstr) == 0) {
                 ; // do nothing
             }
+            else if (strcasecmp("COORD_DIMS_OUT", optstr) == 0) {
+                ; // do nothing
+            }
```
jhamman: You'll never get to this `else if` because of the lines 780-782.  Which do you want?

vic/drivers/image/src/get_global_param.c
```diff
@@ -777,6 +777,9 @@ get_global_param(FILE *gp)
             else if (strcasecmp("SKIPYEAR", optstr) == 0) {
                 sscanf(cmdstr, "%*s %hu", &global_param.skipyear);
             }
+            else if (strcasecmp("COORD_DIMS_OUT", optstr) == 0) {
+                sscanf(cmdstr, "%*s %zu", &options.COORD_DIMS_OUT);
+            }
```
jhamman: see comment at 846.

vic/drivers/image/src/get_global_param.c
```diff
@@ -1102,10 +1111,11 @@ get_global_param(FILE *gp)
...
     else if (global_param.startsec > SEC_PER_DAY) {
-        log_err("The specified simulation start second (%u) > 86400  Make sure "
-                "that the global file defines a positive integer "
-                "for STARTSEC.",
-                global_param.startsec);
+        log_err(
+            "The specified simulation start second (%u) > 86400  Make sure "
+            "that the global file defines a positive integer "
+            "for STARTSEC.",
+            global_param.startsec);
```
jhamman: again, lots of white space changes that don't seem to be necessary or in line with the typical formatting behavior of uncrustify.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -160,12 +174,12 @@ initialize_history_file(nc_file_struct *nc)
...
     }
 
-    status = nc_def_dim(nc->nc_id, "ni", nc->ni_size, &(nc->ni_dimid));
+    status = nc_def_dim(nc->nc_id, "lon", nc->ni_size, &(nc->ni_dimid));
     if (status != NC_NOERR) {
         log_err("Error defining ni dimenension in %s", nc->fname);
     }
 
-    status = nc_def_dim(nc->nc_id, "nj", nc->nj_size, &(nc->nj_dimid));
+    status = nc_def_dim(nc->nc_id, "lat", nc->nj_size, &(nc->nj_dimid));
```
jhamman: Let's think of a way not to hard code these.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -193,10 +207,167 @@ initialize_history_file(nc_file_struct *nc)
...
+            global_param.startyear, global_param.startmonth,
+            global_param.startday);
+
+    status = nc_put_att_text(nc->nc_id, time_var_id, "units",
+                             strlen(str), str);
+    if (status != NC_NOERR) {
+        log_err("Error adding attribute in %s", nc->fname);
+    }
+    status = nc_put_att_text(nc->nc_id, time_var_id, "calendar",
+                             strlen("standard"), "standard");
```
jhamman: We store the calendar in the `global_param_struct` so no need to hardcode.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -193,10 +207,167 @@ initialize_history_file(nc_file_struct *nc)
...
+    }
+    if (global_param.output_steps_per_day == 1) {
+        sprintf(strUnit, "days");
+    }
+    else {
+        sprintf(strUnit, "hours");
+    }
+    sprintf(str, "%s since %i-%02i-%02i 00:00:00", strUnit,
+            global_param.startyear, global_param.startmonth,
+            global_param.startday);
```
jhamman: I'd prefer to hardcode the start date/time to `"days since 0001-01-01 00:00:00"`.  This should work for all calendars and means the time origin is always the same.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -193,10 +207,167 @@ initialize_history_file(nc_file_struct *nc)
...
+    status = nc_put_att_text(nc->nc_id, time_var_id, "standard_name",
+                             strlen("time"), "time");
+    if (status != NC_NOERR) {
+        log_err("Error adding attribute in %s", nc->fname);
+    }
+    if (global_param.output_steps_per_day == 1) {
+        sprintf(strUnit, "days");
+    }
+    else {
+        sprintf(strUnit, "hours");
```
jhamman: get this from `global_param.time_units`.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -193,10 +207,167 @@ initialize_history_file(nc_file_struct *nc)
...
+        status = nc_put_vara_double(nc->nc_id, lat_var_id, dstart, dcount, dvar);
+        if (status != NC_NOERR) {
+            log_err("Error adding data to lat in %s", nc->fname);
+        }
+
+        free(dvar);
+    }
+    else {
+        log_err("COORD_DIMS_OUT should be 1 or 2");
+    }
```
jhamman: you never fill in the time variable.

vic/drivers/image/src/vic_write.c
```diff
@@ -35,92 +35,131 @@
...
+         1) %
+        (global_param.model_steps_per_day /
+         global_param.output_steps_per_day) ==
+        0) {
+        // get start location in NetCDF file for current timestep.
+        nc_time =
+            ((current +
+              1) /
+             (global_param.model_steps_per_day /
+         global_param.output_steps_per_day)) - 1;
```
jhamman: I think it would be better, for now, to use the `date2num` and the current `dmy_struct` to come up with the time value.

vic/drivers/image/src/vic_write.c
```diff
@@ -35,92 +35,131 @@
...
+        dimids[0] = nc_hist_file.time_dimid;
+        dstart[0] = nc_time;
+        dcount[0] = 1;
+
+        if (global_param.output_steps_per_day == 1) {
+            dvar[0] = (double) (current + 1) / 24 - 1;
+        }
+        else {
+            dvar[0] = (double) current;
+        }
```
jhamman: see comment above re. `date2num`.

vic/drivers/image/src/get_global_param.c
jhamman: I'm also wondering if we can just determine this from the domain file.  It seems like that would be easier than making users define it in the global parameter file.

In the global parameter file, users should set the lat/lon coordinate names (e.g. xc or lon and yc or lat).  We should then be able to find out the shape/size of those variables and their dimensions.

jhamman:
@wietsefranssen - good start here.  I made a bunch of comments to keep this moving.

vic/drivers/image/src/get_global_domain.c
```diff
@@ -83,6 +83,18 @@ get_global_domain(char          *nc_name,
         initialize_location(&(global_domain->locations[i]));
     }
 
+    // if MASTER_PROC
+    global_domain->locations_grid = (location_grid_struct *)
+                                    malloc(
+        global_domain->n_ny * global_domain->n_nx *
+        sizeof(location_grid_struct));
```
jhamman: see #257 and update the cast syntax.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -193,10 +207,167 @@ initialize_history_file(nc_file_struct *nc)
...
 
     // leave define mode
     status = nc_enddef(nc->nc_id);
     if (status != NC_NOERR) {
         log_err("Error leaving define mode for %s", nc->fname);
     }
+
+    // fill the netcdf variables lat/lon
+    if (options.COORD_DIMS_OUT == 1) {
+        dvar = (double *) malloc(nc->ni_size * sizeof(double));
```
jhamman: see #257 and update the cast syntax.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -193,10 +207,167 @@ initialize_history_file(nc_file_struct *nc)
...
+        for (i = 0; i < nc->ni_size; i++) {
+            dvar[i] = (double) global_domain.locations_grid[i].longitude;
+        }
+        status = nc_put_vara_double(nc->nc_id, lon_var_id, dstart, dcount, dvar);
+        if (status != NC_NOERR) {
+            log_err("Error adding data to lon in %s", nc->fname);
+        }
+        free(dvar);
+
+        dvar = (double *) malloc(nc->nj_size * sizeof(double));
```
jhamman: see #257 and update the cast syntax.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -193,10 +207,167 @@ initialize_history_file(nc_file_struct *nc)
...
+        }
+
+        status = nc_put_vara_double(nc->nc_id, lat_var_id, dstart, dcount, dvar);
+        if (status != NC_NOERR) {
+            log_err("Error adding data to lon in %s", nc->fname);
+        }
+        free(dvar);
+    }
+    else if (options.COORD_DIMS_OUT == 2) {
+        dvar = (double *) malloc(nc->nj_size * nc->ni_size * sizeof(double));
```
jhamman: see #257 and update the cast syntax.

vic/drivers/image/src/vic_write.c
```diff
@@ -35,92 +35,131 @@
...
-        dstart[i] = -1;
-        dcount[i] = -1;
-    }
+        grid_size = global_domain.n_ny * global_domain.n_nx;
 
-    for (k = 0; k < N_OUTVAR_TYPES; k++) {
-        if (!nc_vars[k].nc_write) {
-            continue;
+        // allocate memory for variables to be stored
+        dvar = (double *) malloc(local_domain.ncells * sizeof(double));
```
jhamman: see #257 and update the cast syntax.

vic/drivers/image/include/vic_driver_image.h
wietsefranssen: That is possible, but keep in mind that in that case the indexes are changing. I mean the following:
In the case you suggest, `location_struct` is bigger because the unmasked gridcells are now included. This means that `location_struct[n]` refers to another lonlat combination. We have to be very careful that this will not effect other parts of the VIC code.
What shall we do now to solve it?

vic/drivers/image/include/vic_driver_image.h
wietsefranssen: Yes I used uncrustify together with `./tools/code_format/uncrustify_VIC_c.cfg`. I also found it unnecessary...

vic/drivers/image/src/get_global_param.c
wietsefranssen: Yes you're right! I removed 844-846. I tested it and its working :)

vic/drivers/image/src/get_global_param.c
wietsefranssen: Sorry I don't understand what you mean. Do you mean the usage of COORD_DIMS_OUT in the global_param file? I remember we talked about it when we were discussing the conventions. We then agreed implement it in the global_param file. Do you have any suggestions about how to get this info from the domain file (how many dimensions the output of the lat/lon values should have)?
Just to explain the statement COORD_DIMS_OUT:
COORD_DIMS_OUT = 1 gives 1D latlon values in the output files. ncdump example of an output file: 
        double lon(lon) ;
                lon:long_name = "longitude" ;
                lon:units = "degrees_east" ;
                lon:standard_name = "longitude" ;
        double lat(lon) ;
                lat:long_name = "latitude" ;
                lat:units = "degrees_north" ;
                lat:standard_name = "latitude" ;

COORD_DIMS_OUT = 2 gives 2D latlon values in the output files. ncdump example of an output file: 
        double lon(lat, lon) ;
                lon:long_name = "longitude" ;
                lon:units = "degrees_east" ;
                lon:standard_name = "longitude" ;
        double lat(lat, lon) ;
                lat:long_name = "latitude" ;
                lat:units = "degrees_north" ;
                lat:standard_name = "latitude" ;

wietsefranssen: About the coordinate names: I know about those plans. Shall we implement this?

vic/drivers/image/src/get_global_param.c
wietsefranssen: see: https://github.com/UW-Hydro/VIC/pull/303/files#r44461028

vic/drivers/image/src/vic_init_output.c
wietsefranssen: Suggestion: Let the user choose from the global_param file. Shall we implement this?

vic/drivers/image/src/vic_init_output.c
wietsefranssen: We will have a look at it.

vic/drivers/image/src/vic_init_output.c
wietsefranssen: We don't understand why this should give problem. It is in fact the time which is extracted from the global_param and provided by the user.
I expect some problem with analysing software when using this date. An other minor detail is that "0001-01-01 00:00:00" can be confusing for users (users maybe thing it is not a real date).
A date like 1900-01-01 00:00:00 could also be used (it is more intuitive).

vic/drivers/image/src/vic_init_output.c
wietsefranssen: The point is you need to specify this in a attribute as a non-capital string (in order follow the netcdf conventions).
A way to avoid it is to always use "hours". The downside of it is that it can be confusing to have daily output with the string "hours" in the Netcdf output.

vic/drivers/image/src/get_global_domain.c
wietsefranssen: I changed it into (made also a calloc of it):
     `global_domain->locations_grid = calloc(
        global_domain->n_ny * global_domain->n_nx, sizeof(*global_domain->locations_grid));`
I just used the same syntax as a few lines earlier eg line 76. Why do these one still use the old cast syntax? Or just because you didn't push the last commit to github yet?

vic/drivers/image/include/vic_driver_image.h
jhamman: out of curiosity, what version of `uncrustify` are you using?  @bartnijssen and I using `0.60`.

wietsefranssen: 0.59!
I will update my version to 0.60 and try it again. I will let you know tomorrow. Still weird :)

vic/drivers/image/src/get_global_param.c
jhamman: > About the coordinate names: I know about those plans. Shall we implement this?

I think that makes sense, since you are already in the neighborhood.

>Do you have any suggestions about how to get this info from the domain file (how many dimensions the output of the lat/lon values should have)

`nc_inq_var` should do it.  In [this](https://www.unidata.ucar.edu/software/netcdf/docs/netcdf-c/nc_005finq_005fvar.html#nc_005finq_005fvar) example, the number of dimensions (`rh_ndims`) is determined in the second to last line.

vic/drivers/image/src/vic_init_output.c
jhamman: Yes, since you are already in there, why don't you give it a try.

vic/drivers/image/src/vic_init_output.c
jhamman: right, you'll need to convert the integer stored in `global_param.output_steps_per_day` to a lowercase string but we should decide which units to use based on `global_param.output_steps_per_day`.  The definition of the integers stored therein is here: https://github.com/UW-Hydro/VIC/blob/develop/vic/vic_run/include/vic_def.h#L525-L531

vic/drivers/image/src/get_global_domain.c
jhamman: That PR is still in the works (see #257).

vic/drivers/image/src/get_global_param.c
wietsefranssen: I know nc_inq_var but thats not what I mean. The 1d or 2d info it still not in the domain. Do you have time tomorrow (same time as now) to discuss this on gitter?

vic/drivers/image/src/vic_init_output.c
wietsefranssen: I have seen that but this is not "days" or "hours" or something like that. The convention needs exactly those words (analysis software needs that as well). Lets discuss this soon on gitter.

vic/drivers/image/src/get_global_domain.c
wietsefranssen: Okay I'll wait for a next commit :)

vic/drivers/image/src/vic_init_output.c
jhamman: so the logic should go something like this:

```C
if (global_param.output_steps_per_day == TIME_UNITS_DAYS) {
    sprintf(strUnit, "days");
}
else if (global_param.output_steps_per_day == TIME_UNITS_HOURS) {
    sprintf(strUnit, "hours");
}
...

File: vic/drivers/image/src/vic_init_output.c (hunk 7 above)
Comment (jhamman) [inline vic/drivers/image/src/vic_init_output.c]: the logic will be similar to the time units example code I posted above.

File: vic/drivers/image/src/vic_init_output.c (hunk 8 above)
Comment (jhamman) [inline vic/drivers/image/src/vic_init_output.c]: This date doesn't need to be intuitive or even seem like a real date.  It only needs to be a valid date in the calendar (which "0001-01-01" for all calendars). Most users will not interact with this date string at all, it is most frequently used by software that interprets the time variable.  Choosing "0001-01-01" has two main advantages:

1) It is the earliest valid date for all calendars
2) It allows VIC to be run before 1900 without producing negative timestamps (I've seen problems before)

"0001-01-01" is pretty commonly used in the climate modeling community.

Comment (wietsefranssen) [inline vic/drivers/image/src/vic_init_output.c]: I have had many people getting confused by it. But personally I don't mind.
Lets discuss this further on gitter as well (because I also want to know why you prefer to hardcode it in stead of using the VIC start time).

File: vic/drivers/image/src/vic_init_output.c (hunk 9 above)
Comment (wietsefranssen) [inline vic/drivers/image/src/vic_init_output.c]: Ah thats what you mean, I understood it different. Thats indeed better I will change it accordingly.

**Comment by bartnijssen** [comment]
@wietsefranssen : Can you resolve the merge conflicts before I merge this? Since you started this branch a while ago, you probably need to merge the existing develop into your branch first. After that you probably won't have any merge conflicts.

**Comment by jhamman** [comment]
@bartnijssen - I don't think we're ready to merge here yet. We still need to resolve the calendar and time handling pieces.

**Comment by bartnijssen** [comment]
OK - sounds good - still need to resolve the conflict though.

File: vic/drivers/image/src/vic_init_output.c [hunk 18]
```diff
@@ -193,10 +208,212 @@ initialize_history_file(nc_file_struct *nc)
...
+        sprintf(strUnit, "minutes");
+    }
+    else if (global_param.time_units == TIME_UNITS_HOURS) {
+        sprintf(strUnit, "hours");
+    }
+    else if (global_param.time_units == TIME_UNITS_DAYS) {
+        sprintf(strUnit, "days");
+    }
+    else {
+        sprintf(strUnit, "-");
```
jhamman: better to raise an error here if we don't know what the time units are.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -193,10 +208,212 @@ initialize_history_file(nc_file_struct *nc)
...
+        sprintf(strCalendar, "julian");
+    }
+    else if (global_param.calendar == CALENDAR_ALL_LEAP) {
+        sprintf(strCalendar, "all leap");
+    }
+    else if (global_param.calendar == CALENDAR_366_DAY) {
+        sprintf(strCalendar, "366 day");
+    }
+    else {
+        sprintf(strCalendar, "no calendar specified");
```
jhamman: raise an error here.  It is possible to not have a calendar attribute in the time variable but I don't think the current `implementation of `vic_time` can handle that.

vic/drivers/image/src/vic_write.c
```diff
@@ -35,26 +35,29 @@
...
+    double                    *dvar = NULL;
+    size_t                     dcount[MAXDIMS];
+    size_t                     dstart[MAXDIMS];
+    int                        nc_time;
 
     grid_size = global_domain.n_ny * global_domain.n_nx;
 
     // allocate memory for variables to be stored
-    dvar = (double *) malloc(local_domain.ncells * sizeof(double));
+    dvar = calloc(local_domain.ncells, sizeof(*dvar));
```
jhamman: did you mean to change this from `malloc` to `calloc`?

wietsefranssen: Yes, but it was a mistake of me while adapting the cast syntax you mentioned. I am used to use calloc in stead of malloc (calloc is a bit slower because it it zeroing out memory is'nt it?). anyway I changed it to:
```
    dvar = malloc(local_domain.ncells * sizeof(*dvar));
```

vic/drivers/image/include/vic_driver_image.h
jhamman: Okay, let's leave this as is here and we'll revisit it in #313.

vic/drivers/image/src/get_global_domain.c
```diff
@@ -186,6 +207,16 @@ initialize_location(location_struct *location)
 }
 
 /******************************************************************************
+ * @brief    Initialize location_grid structure.
+ *****************************************************************************/
+void
+initialize_location_grid(location_grid_struct *location_grid)
+{
+    location_grid->latitude = 0;
+    location_grid->longitude = 0;
```
jhamman: can you use `MISSING` instead of 0 here?

vic/drivers/image/src/get_global_param.c
jhamman: are we clear on this?  just check to see if the latitude/longitude vars are 1d or 2d.

vic/drivers/image/src/get_global_param.c
jhamman: You may need to manually undo all these whitespace changes.

vic/vic_run/include/vic_def.h
```diff
@@ -342,6 +342,9 @@ typedef struct {
     bool COMPRESS;       /**< TRUE = Compress all output files */
     bool MOISTFRACT;     /**< TRUE = output soil moisture as fractional moisture content */
     size_t Noutfiles;    /**< Number of output files (not including state files) */
+    size_t COORD_DIMS_OUT;    /**< Number of output dimensions for lat and lon variables. COORD_DIMS_OUT 1 = lon(lon), lat(lat). COORD_DIMS_OUT 2 = lon(xc, yc), lat(xc, yc) */
+    char DOMAIN_LON_VAR[MAXSTRING];    /**< Name of the variable and the dimension of longitude in the output file */
+    char DOMAIN_LAT_VAR[MAXSTRING];    /**< Name of the variable and the dimension of latitude in the output file */
```
jhamman: These need to be added to the `options_struct` in `vic_mpi_support.c`

jhamman:
@wietsefranssen - almost there.  I'm working on #312 today.

wietsefranssen:
@jhamman Hi Joe, I just made some changes in order to retreive the number of dimensions from the domain file. I also removed the `global_domain_grid` struct and replaced them for `ncells_total`.
It is still perfect yet (need to tidy up and need fix one mpi error).
In case you find some time, it would be nice to have a quick look at it to give some suggestions.

vic/drivers/image/src/get_nc_varndimensions.c
```diff
@@ -0,0 +1,69 @@
...
+#include <vic_def.h>
+#include <vic_run.h>
+#include <vic_driver_image.h>
+
+/******************************************************************************
+ * @brief    Get netCDF dimension.
+ *****************************************************************************/
+int
+get_nc_varndimensions(char *nc_name,
+                 char *var_name)
```
jhamman: the indentation looks off here.

vic/drivers/image/src/vic_init_output.c
```diff
@@ -160,12 +175,16 @@ initialize_history_file(nc_file_struct *nc)
         log_err("Error defining nlayer dimenension in %s", nc->fname);
     }
 
-    status = nc_def_dim(nc->nc_id, "ni", nc->ni_size, &(nc->ni_dimid));
+    status =
+        nc_def_dim(nc->nc_id, options.DOMAIN_LON_VAR, nc->ni_size,
+                   &(nc->ni_dimid));
```
jhamman: I think we need to do a conditional here for the number or coordinate dims
1) if `coord_dims == 1`, then the dim names are `lon/lat`
2) if `coord_dims == 2`, then the dim names are `ni/nj`

vic/vic_run/include/vic_def.h
```diff
@@ -60,7 +60,7 @@
 #define MIN_VEGCOVER    0.0001 /**< Minimum allowable vegcover fraction */
 
 /***** Define minimum and maximum values for model timesteps *****/
-#define MIN_SUBDAILY_STEPS_PER_DAY  4
+#define MIN_SUBDAILY_STEPS_PER_DAY  1
```
jhamman: this we do not want.  Because we have removed `mtclim` in the image driver, you may run into a bunch of problems here.

jhamman:
This looks pretty good.  I had a few minor comments.  Once you've addressed those, please ping @bartnijssen and @tbohn for their code reviews.

vic/vic_run/include/vic_def.h
wietsefranssen: Ah it was a mistake from me to commit this line. However we need to talk about it because I still cannot run on a daily basis (subdaily timestep > 1) without wrong output.
Skype this evening (morning for you :))?

jhamman: I don't think I understand what you are trying to do.  The VIC image driver will not support a daily timestep. You will need to supply forcings at a maximum timestep of 6 hours. Is there a particular reason you want to run VIC at a daily timestep?

Previous versions of VIC disaggregated daily met forcings to subdaily on the fly using MTCLIM, but this cannot be done in any reasonable way in the image driver. We have also removed MTCLIM from the classic driver in VIC 5.  Relevant closed issues are #7 and #240.

wietsefranssen:
@jhamman I just fixed the things Joe suggested. @tbohn and @bartnijssen Can you please have a look at the code? Thank you beforehand.

jhamman:
@tbohn and @bartnijssen - last call for comments here.

@wietsefranssen - can you merge in the current `develop` branch?

I'll give this one last read through and then merge if there are no failing tests and no objections by others.

jhamman:
this pr closes #297 and closes #197

jhamman:
Merging now. Thanks @wietsefranssen.

Issue #306, opened 2015-11-13
Veg_hist code in 4.2 lacks an update


tbohn:
I have discovered that my original implementation of the veg_hist feature in release 4.2 was missing a modification I subsequently made in my private version but forgot to add to GIT.  This modification results in minor changes to soil evaporation magnitudes but is necessary to make 4.2 match the results and description in my manuscript.  Changes are confined to two files: arno_evap.c and func_surf_energy_bal.c.

I'd like to add this as a hotfix to 4.2.  Would that be OK?

tbohn:
The code changes for this are ready on my own branch off master, and I can make a pull request, but I don't know which official branch to make the pull request to.

jhamman:
I think 4.2.c is fine.  I can backport to 5.  Make a PR against master.

PR #309, opened 2015-11-17
Update to code for Esoil in sparse vegetation.

tbohn:
I had intended to add this update to the original 4.2 release, but forgot about it until now.  This addresses issue #306 .

### DIFF
--- a/src/arno_evap.c
+++ b/src/arno_evap.c
@@ -85,6 +85,7 @@ double arno_evap(layer_data_struct *layer,
   double max_infil;
   double Evap;
   double tmpsum;
+  double rarc_soil = 100.0;
 
   Evap = 0;
 
@@ -97,7 +98,7 @@ double arno_evap(layer_data_struct *layer,
 
   /* Calculate the potential bare soil evaporation (mm/time step) */
   
-  Epot = penman(air_temp, elevation, rad, vpd, ra, 0.0, 0.0) * delta_t / SEC_PER_DAY;
+  Epot = penman(air_temp, elevation, rad, vpd, ra, 0.0, rarc_soil) * delta_t / SEC_PER_DAY;
   
   /**********************************************************************/
   /*  Compute temporary infiltration rate based on given soil_moist.    */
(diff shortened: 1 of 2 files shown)


bartnijssen:
@jhamman : Do you want to include this into the master directly or use a hot fix branch where we can run some tests first?

src/arno_evap.c
```diff
@@ -85,6 +85,7 @@ double arno_evap(layer_data_struct *layer,
   double max_infil;
   double Evap;
   double tmpsum;
+  double rarc_soil = 100.0;
```
jhamman: Why `100.0`?

src/func_surf_energy_bal.c
```diff
@@ -688,6 +691,32 @@ double func_surf_energy_bal(double Ts, va_list ap)
...
+      }
+      else {
+        Ra_used[0] = 0;
+      }
+    }
+    else {
+      Ra_used[0] = 0;
+    }
+  }
+
```
jhamman: Can you add some comments here.  There are a few hard coded literals and a lot of moving pieces without any explanation.

src/arno_evap.c
tbohn: Gives better results than 0 or 200.  I didn't do a fancy Bayesian parameter
estimation, not enough time for that.

I apologize for making this a hard-wired value - I considered making it
another input soil parameter, but we have so many of those already.  And
this likely won't noticeably affect anyone who has been using the
traditional VIC set-up, since bare soil rarely occurs in the traditional
parameters.

jhamman:
Also @tbohn, can you redirect this PR to the [hotfix.4.2.c](https://github.com/UW-Hydro/VIC/tree/hotfix/4.2.c) branch.

tbohn:
Is there a way to change the target branch without closing this PR and opening a new one?  I searched on google and found a GitHub issue that's been open for 2 years on this very subject - tons of users agree that it would be great to be able to change the target branch, but Git apparently hasn't done it yet...  So, I suppose I'll close this and open a new one.

jhamman:
Yeah, I think you'll need to close the PRs.  Sorry for the misdirection early on.

PR #315, opened 2015-11-30
Update to code for Esoil in sparse vegetation.

tbohn:
I had intended to add this update to the original 4.2 release, but forgot about it until now.  This addresses issue #306.

### DIFF
--- a/src/arno_evap.c
+++ b/src/arno_evap.c
@@ -85,6 +85,7 @@ double arno_evap(layer_data_struct *layer,
   double max_infil;
   double Evap;
   double tmpsum;
+  double rarc_soil = 100.0;
 
   Evap = 0;
 
@@ -97,7 +98,7 @@ double arno_evap(layer_data_struct *layer,
 
   /* Calculate the potential bare soil evaporation (mm/time step) */
   
-  Epot = penman(air_temp, elevation, rad, vpd, ra, 0.0, 0.0) * delta_t / SEC_PER_DAY;
+  Epot = penman(air_temp, elevation, rad, vpd, ra, 0.0, rarc_soil) * delta_t / SEC_PER_DAY;
   
   /**********************************************************************/
   /*  Compute temporary infiltration rate based on given soil_moist.    */
--- a/src/arno_evap.c
+++ b/src/arno_evap.c
@@ -85,9 +85,6 @@ double arno_evap(layer_data_struct *layer,
   double max_infil;
   double Evap;
   double tmpsum;
-  /* best-fitting value of architectural resistance, determined from
-     evaluation of model performance at eddy covariane tower sites */
-  double rarc_soil = 100.0; // [s/m]
 
   Evap = 0;
 
@@ -100,7 +97,7 @@ double arno_evap(layer_data_struct *layer,
 
   /* Calculate the potential bare soil evaporation (mm/time step) */
   
-  Epot = penman(air_temp, elevation, rad, vpd, ra, 0.0, rarc_soil) * delta_t / SEC_PER_DAY;
+  Epot = penman(air_temp, elevation, rad, vpd, ra, 0.0, RARC_SOIL) * delta_t / SEC_PER_DAY;
   
   /**********************************************************************/
   /*  Compute temporary infiltration rate based on given soil_moist.    */
--- a/src/arno_evap.c
+++ b/src/arno_evap.c
@@ -85,7 +85,9 @@ double arno_evap(layer_data_struct *layer,
   double max_infil;
   double Evap;
   double tmpsum;
-  double rarc_soil = 100.0;
+  /* best-fitting value of architectural resistance, determined from
+     evaluation of model performance at eddy covariane tower sites */
+  double rarc_soil = 100.0; // [s/m]
 
   Evap = 0;
 
--- a/src/func_surf_energy_bal.c
+++ b/src/func_surf_energy_bal.c
@@ -300,6 +300,9 @@ double func_surf_energy_bal(double Ts, va_list ap)
   double tmp_displacement[3];
   double tmp_roughness[3];
   double tmp_ref_height[3];
+  double ga_veg;
+  double ga_bare;
+  double ga_average;
 
   /************************************
     Read variables from variable list 
@@ -688,6 +691,32 @@ double func_surf_energy_bal(double Ts, va_list ap)
   else
     Ra_used[0] = HUGE_RESIST;
   
+  if (veg_var->vegcover < 1) {
+    if (Ra_used[0] > 0) {
+      ga_veg = 1/Ra_used[0];
+      tmp_wind[0] = wind[0];
+      tmp_wind[1] = -999.;
+      tmp_wind[2] = -999.;
+      tmp_height = soil_con->rough/0.123;
+      tmp_displacement[0] = calc_veg_displacement(tmp_height);
+      tmp_roughness[0] = soil_con->rough;
+      tmp_ref_height[0] = 10;
+      Error = CalcAerodynamic(0,0,0,soil_con->snow_rough,soil_con->rough,0,Ra_bare,tmp_wind,tmp_displacement,tmp_ref_height,tmp_roughness);
+      Ra_bare[0] /= StabilityCorrection(tmp_ref_height[0], tmp_displacement[0], TMean, Tair, tmp_wind[0], tmp_roughness[0]);
+      if (Ra_bare[0] > 0) {
+        ga_bare = 1/Ra_bare[0];
+        ga_average = veg_var->vegcover*ga_veg + (1-veg_var->vegcover)*ga_bare;
+        Ra_used[0] = 1/ga_average;
+      }
+      else {
+        Ra_used[0] = 0;
+      }
+    }
+    else {
+      Ra_used[0] = 0;
+    }
+  }
+
   /*************************************************
     Compute Evapotranspiration if not snow covered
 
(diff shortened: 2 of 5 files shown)


jhamman:
@tbohn - can you address the comments in #309?

src/arno_evap.c
```diff
@@ -85,6 +85,7 @@ double arno_evap(layer_data_struct *layer,
   double max_infil;
   double Evap;
   double tmpsum;
+  double rarc_soil = 100.0;
```
tbohn: @jhamman commented: why 100.0?
@tbohn's reply: 100 gives better results than 0 or 200, from tests I conducted at eddy covariance towers. I didn't do a fancy Bayesian parameter estimation, not enough time for that. I apologize for making this a hard-wired value - I considered making it another input soil parameter, but we have so many of those already. And this likely won't noticeably affect anyone who has been using the traditional VIC set-up, since bare soil rarely occurs in the traditional parameters.

src/func_surf_energy_bal.c
```diff
@@ -688,6 +691,32 @@ double func_surf_energy_bal(double Ts, va_list ap)
...
+      }
+      else {
+        Ra_used[0] = 0;
+      }
+    }
+    else {
+      Ra_used[0] = 0;
+    }
+  }
+
```
tbohn: @jhamman commented: Can you add some comments here. There are a few hard coded literals and a lot of moving pieces without any explanation.
@tbohn's reply: OK, will do.

src/arno_evap.c
```diff
@@ -85,6 +85,9 @@ double arno_evap(layer_data_struct *layer,
   double max_infil;
   double Evap;
   double tmpsum;
+  /* best-fitting value of architectural resistance, determined from
+     evaluation of model performance at eddy covariane tower sites */
+  double rarc_soil = 100.0; // [s/m]
```
jhamman: I still don't love this but if @bartnijssen is okay with it, we can keep it for now.  Maybe we should `#define` this in `vicNl_def.h`?

When we port these changes to VIC.5, `rarc_soil` needs to go into the `parameters_struct` or into the `soil_con_struct`.

jhamman:
@bartnijssen - ready for you.

src/func_surf_energy_bal.c
```diff
@@ -687,15 +690,65 @@ double func_surf_energy_bal(double Ts, va_list ap)
...
+    /** If Ra_used (which pertains to under vegetation at this point)
+        is non-zero, use it to compute area-weighted average **/
+    if (Ra_used[0] > 0) {
+      /** aerodynamic conductance under vegetation **/
+      ga_veg = 1/Ra_used[0];
+      /** compute aerodynamic resistance over exposed soil (Ra_bare) **/
+      tmp_wind[0] = wind[0];
+      tmp_wind[1] = -999.; // unused
+      tmp_wind[2] = -999.; // unused
+      tmp_height = soil_con->rough/0.123; // eqn from calc_veg_roughness()
```
bartnijssen: No hardcoded values - if this is a constant, it needs to be set in one of the header files.

src/func_surf_energy_bal.c
```diff
@@ -687,15 +690,65 @@ double func_surf_energy_bal(double Ts, va_list ap)
...
+      tmp_displacement[0] = calc_veg_displacement(tmp_height);
+      tmp_roughness[0] = soil_con->rough;
+      tmp_ref_height[0] = 10; // wind measurement height over bare soil
```
bartnijssen: No hardcoded values (see above)

src/func_surf_energy_bal.c
```diff
@@ -687,15 +690,65 @@ double func_surf_energy_bal(double Ts, va_list ap)
...
+        /** aerodynamic conductance over exposed soil **/
+        ga_bare = 1/Ra_bare[0];
+        /** area-weighted average aerodynamic conductance **/
+        ga_average = veg_var->vegcover*ga_veg + (1-veg_var->vegcover)*ga_bare;
+        /** aerodynamic resistance is inverse of conductance **/
+        Ra_used[0] = 1/ga_average;
+      }
+      /** else aerodynamic resistance is zero **/
+      else {
+        Ra_used[0] = 0;
```
bartnijssen: Should this be zero or infinite (or does it not matter because it means that the value is not used?)

src/func_surf_energy_bal.c
bartnijssen: Still needs those comments. The hardcoded values need to be moved to header files.

bartnijssen:
@tbohn: Can you show what the results of this change is?

@jhamman: Go ahead and merge after comments are addressed.

src/func_surf_energy_bal.c
tbohn: But I added several lines of comments...  It needs *more*?  Which lines in particular need comments?

src/func_surf_energy_bal.c
tbohn: It should be zero.  Ra of 0 means infinite conductance.  Perhaps some of the confusion here is due to the naming of the variables.  Before I added this feature, Ra_used was the Ra of the entire veg tile.  Since vegcover was assumed == 1.0 in the past, Ra_used was also the Ra of the vegetated portion (which was 100% of the tile).  When I added this block, I intended to adjust Ra_used for the presence of exposed soil.  In this case, the old Ra_used now initially only pertains to the vegetated portion (which is < 100% of the tile), and gets reassigned to the value pertaining to the entire tile later on.

So, what we have is:
If vegetated Ra is 0, then vegetated conductance (ga) is infinite, so total tile ga is also infinite, so Ra_used = 1/ga is 0.
If exposed soil Ra is 0, Ra_used also = 0.

What if I add another intermediate variable to make this clearer: Ra_veg.  Ra_veg will be assigned = Ra_used at the beginning of this block.  Then, ga_veg = 1/Ra_veg.  ga_avg = weighted average as currently stated.  Then, Ra_used = 1/ga_avg.  This will hopefully make it clearer that Ra_used = 0 if either Ra_veg or Ra_bare are 0.

bartnijssen: Yes - that would be much easier to understand - please go ahead.

src/func_surf_energy_bal.c
tbohn: OK, but release 4.2 has a bunch of hardcoded values throughout the code.  For example, the 0.123 is also used as a hardcoded value in calc_veg_roughness(); 0.67 is used in calc_veg_displacement() and calc_veg_height(); the -999s are used in full_energy() (I took these lines from full_energy()).  I can replace all of these with constants defined in vicNl_def.h, both in this function and in the others that I just listed.  But does this really need to be done in 4.2, when constants are overhauled in 5.0?

tbohn:
How should I show the results - should I attach a plot of ET before/after?

src/arno_evap.c
tbohn: OK, I'm defining a constant RARC_SOIL in vicNl_def.h...

src/func_surf_energy_bal.c
tbohn: OK, I'm going to replace the -999s with ERROR (which is defined as -999 in vicNl_def.h), both here and in full_energy(), which is where I copied these lines from.

I'm defining a few other constants too:
RATIO_DH_HEIGHT_VEG 0.67 - used in calc_veg_displacement() and calc_veg_height()
RATIO_RL_HEIGHT_VEG 0.123 - used both here and in calc_veg_roughness()
WINDH_SOIL 10.0

src/func_surf_energy_bal.c
tbohn: I'm just going ahead and defining these constants in vicNl_def.h and replacing the numbers here with those constants.

src/func_surf_energy_bal.c
tbohn: OK, done.

jhamman:
@tbohn - a plot of ET separate plots of E and T would be great.

tbohn:
First, some performance statistics at eddy covariance towers - mostly Ameriflux but also 5 other towers in Mexico:
Class | N | Old Median NSE | New Median NSE | Change
--------|---|----------------------|-----------------------|------------
ENF | 8 | 0.53 | 0.51 | -0.02
DBF | 2 | 0.61 | 0.57 | -0.04
CSH | 4 | 0.23 | 0.25 | +0.02
OSH | 8 | 0.61 | 0.62 | +0.01
WSA | 2 | 0.40 | 0.49 | +0.09
SAV | 1 | 0.50 | 0.53 | +0.03
GRA | 21 | 0.38 | 0.55 | +0.17

(ENF = Evergreen Needleleaf Forest; DBF = Deciduous Broadleaf Forest; CSH = Closed Shrubland; OSH = Open Shrubland; WSA = Woody Savanna; SAV = Savanna; GRA = Grassland)

So, you can see that changes are minimal in forest biomes, but a big improvement in more sparsely-vegetated biomes such as savannas and grasslands.

jhamman:
(I fixed you table - nice try though :wink: )

tbohn:
Ah, so I just needed to insert a blank line between the non-table text and table text?

tbohn:
Here's an example of before and after the change, at the Santa Rita Mesquite Ameriflux Tower (Open Shrubland).

Old Model (NSE = 0.64):
plot et srm 2008 fix_ra_test old_model

New Model (NSE = 0.73):
plot et srm 2008 fix_ra_test new_model

As you can see, the main effect is to reduce Esoil a bit (primarily due to the RARC_SOIL of 100 s/m).

Here's another example, from Diablo (Grassland).

Old Model (NSE = 0.32):
plot et dia 2011 fix_ra_test old_model

New Model (NSE = 0.53):
plot et dia 2011 fix_ra_test new_model

I can include plots of other sites if you'd like to see more...

tbohn:
@jhamman and @bartnijssen - do you need me to provide any more test results for this?

jhamman:
This looks good to me.  Thanks for showing some results @tbohn.  @bartnijssen?

bartnijssen:
@tbohn: Thanks for the plots and explanations

Issue #308, opened 2015-11-17
Water balance error in lake model


tbohn:
The lake model in release 4.2 is exhibiting water balance errors.  Currently investigating to find the source of the problem.

So far, I have found at least one part of the problem: the variable out_prec, used only for reporting precipitation for output, was not being updated correctly in full_energy.c.  This only affects the water balance check in put_data.c (as well as making the output precipitation not equal the input precipitation).  This is easy to fix.  However, after fixing this, there are still water balance errors, which I am investigating now.

tbohn:
As it turns out, the OUT_PREC bug is the only code bug.  The other issue was caused by a bad parameter value, for which I have added some extra validation in read_lakeparam.c.

tbohn:
@jhamman this issue can be closed since you've merged pull request #316

PR #310, opened 2015-11-17
Fix to calculation of OUT_PREC for output files when LAKES=TRUE.

tbohn:
This addresses issue #308.

PR #316, opened 2015-11-30
Fix to calculation of OUT_PREC for output files when LAKES=TRUE.

tbohn:
This addresses issue #308.


jhamman:
@tbohn - can you address the comments in #310 here.

tbohn:
There are no outstanding comments to address from #310.  The only code comment was @bartnijssen 's request to run uncrustify, which he then cancelled due to the rest of release 4.2 not having been uncrustified.

bartnijssen:
@jhamman: Go ahead and merge. Please run the files through uncrustify before adding them to VIC 5.

Issue #311, opened 2015-11-19
Buffer overflow when using optimization flags


wietsefranssen:
When you compile VIC with optimization enabled (`CFFLAG -O1` or higher) in stead of `-O0`, VIC will not run.
It throws the following messae after the vic-info text:
```*** buffer overflow detected ***```

note: I only tested this in VIC_image mode.

**Comment by jhamman** [comment]
Can you post the compiler and platform specifications that reproduce this behavior?

**Comment by wietsefranssen** [comment]
```
mpicc for MPICH version 3.0.4
Using built-in specs.
COLLECT_GCC=cc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/4.8/lto-wrapper
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 4.8.4-2ubuntu1~14.04' --with-bugurl=file:///usr/share/doc/gcc-4.8/README.Bugs --enable-languages=c,c++,java,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-4.8 --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --with-gxx-include-dir=/usr/include/c++/4.8 --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --enable-gnu-unique-object --disable-libmudflap --enable-plugin --with-system-zlib --disable-browser-plugin --enable-java-awt=gtk --enable-gtk-cairo --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-4.8-amd64/jre --enable-java-home --with-jvm-root-dir=/usr/lib/jvm/java-1.5.0-gcj-4.8-amd64 --with-jvm-jar-dir=/usr/lib/jvm-exports/java-1.5.0-gcj-4.8-amd64 --with-arch-directory=amd64 --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --enable-objc-gc --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --with-tune=generic --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
[...]

PR #320, opened 2015-12-08
buffer overflow fix

wietsefranssen:
Changed the size of char `varname[20];` to `char varname[30];` in `vic_driver_shared.h`. This has to be done because some strings in `output_list_utils.c` are longer than 20 and this will cause buffer overflows.


jhamman:
xref: #311

bartnijssen:
I'd prefer if we did not use `30`, but rather something like `MAXSTRING`, which is used in the rest of the code for string sizes (defined in `vic_def.h`). I will make the change and open a PR for that.

wietsefranssen:
It was a hardcoded number, so I just changes it to an other one. But I completely agree with that.

Issue #319, opened 2015-12-04
VIC overwrites veg_lib values of LAI, albedo, and vegcover with those of the current cell


tbohn:
The function read_vegparam() overwrites the values of LAI, albedo, and vegcover in the veg_lib structure with those from the veg_param file for the current cell, if they are listed there and if LAI_SRC, ALB_SRC, or VEGCOVER_SRC are set to FROM_VEGPARAM.

In olden times this was not a problem - the veg_lib structure's function was re-purposed to contain climatological values of LAI and albedo for the current cell.  However, the new veg_hist feature in release 4.2 allows for the values of LAI, albedo, and vegcover in the veg_param or veg_hist files to be -1 (the "nodata" flag for these quantities), which was intended to indicate to VIC that it should take the value from the veg_param file if "nodata" in the veg_hist file, or from the veg_lib file if "nodata" in the veg_param file.  The reason for this is that some global datasets of these quantities (e.g. MODIS products) have nodata values in some cells on some dates due to cloud cover or other poor retrievals.

Unfortunately, for this to work properly, we can't have VIC overwriting the veg_lib structure with the current cell's veg_param values.  If we encounter a cell for which some of these values are "nodata", VIC will use what's in veg_lib, which will be the veg_param value from the last cell that had a valid value.

To fix this, I propose we add some 12-element arrays to the veg_con structure, to store the values of LAI, albedo, vegcover, and Wdmax (which depends on LAI) of the current cell without overwriting the veg_lib values.  So, the veg_lib structure will essentially be a read-only copy of the veg_library file in memory.

I propose that this be included in the upcoming hotfix 4.2.c.

tbohn:
Am I correct in assuming you'd prefer that I create my branch off of the official master, and that I should not include the contents of my other 2 (pending) pull requests?

Also, are my 2 pending pull requests waiting on me for anything?  Are they essentially approved or is there some chance you will want me to change something?

jhamman:
branch off master and issue against 4.2.c.

@bartnijssen and I will take another pass through your other prs asap.

jhamman:
@tbohn - responding to this issue.  I think we want to ovoid any time information inside `vic_run`.  In my opinion, the solution here needs to fill a single set of scalar veg params that go into `vic_run`.  However the driver decides to do that is up to the individual driver.

tbohn:
This is just a fix to 4.2... I can find a different fix for 5.0...

tbohn:
[...]

PR #321, opened 2015-12-08
Fix to overwriting of veg_lib with cell-specific values.

tbohn:
This addresses issue #319

### DIFF
--- a/src/read_vegparam.c
+++ b/src/read_vegparam.c
@@ -239,7 +239,7 @@ veg_con_struct *read_vegparam(FILE *vegparam,
         nrerror(ErrStr);
       }
 
-      for ( j = 0; j < 12; j++ ) {
+      for ( j = 0; j < MONTHSPERYEAR; j++ ) {
         temp[i].albedo[j] = veg_lib[temp[i].veg_class].albedo[j];
         temp[i].LAI[j] = veg_lib[temp[i].veg_class].LAI[j];
         temp[i].vegcover[j] = veg_lib[temp[i].veg_class].vegcover[j];
--- a/src/read_vegparam.c
+++ b/src/read_vegparam.c
@@ -239,16 +239,23 @@ veg_con_struct *read_vegparam(FILE *vegparam,
         nrerror(ErrStr);
       }
 
+      for ( j = 0; j < 12; j++ ) {
+        temp[i].albedo[j] = veg_lib[temp[i].veg_class].albedo[j];
+        temp[i].LAI[j] = veg_lib[temp[i].veg_class].LAI[j];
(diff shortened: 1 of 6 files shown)


tbohn:
I tested this to make sure that LAI, albedo, and vegcover were identical to those in the veg_library when -1s were specified for them in the veg_param file.

src/read_vegparam.c
```diff
@@ -239,16 +239,23 @@ veg_con_struct *read_vegparam(FILE *vegparam,
         nrerror(ErrStr);
       }
 
+      for ( j = 0; j < 12; j++ ) {
```
jhamman: MONTHSPERYEAR is defined in `vicNl_def.h` so you might as well use it.

src/vicNl_def.h
```diff
@@ -1016,6 +1016,10 @@ typedef struct {
   float   fetch;            /* Average fetch length for each vegetation class. */
   int     LAKE;             /* TRUE = this tile is a lake/wetland tile */
   double *CanopLayerBnd;    /* Upper boundary of each canopy layer, expressed as fraction of total LAI */
+  double albedo[12];       /* climatological vegetation albedo (fraction) */
+  double LAI[12];          /* climatological leaf area index (m2/m2) */
+  double vegcover[12];     /* climatological fractional area covered by plants within the tile (fraction) */
+  double Wdmax[12];        /* climatological maximum dew holding capacity (mm) */
```
jhamman: Use `MONTHSPERYEAR` in place of all these 12s.

src/read_vegparam.c
tbohn: But there are several other existing loops over j from 0 to 12 in this function.  Changing only this one will be inconsistent with the rest of the function...  And changing all of them, well, is it my job to overhaul all of the constants just because I'm making a rather minor change to the function?

jhamman:
@tbohn - these changes seem pretty benign.  I had just two small comments and then we should be good to go here.

src/read_vegparam.c
jhamman: Answer - yes it is your job.  This will be merged into develop and if its not your job, it quickly becomes mine :).  Just change this one location.

tbohn: OK, I didn't realize that 4.2. was having constants replaced.  I'll do it.  Stay tuned...

jhamman: Its just good practice and we do merge this back into VIC 5, even if we wont be keeping it for long.  So if we don't do it now, I always have to clean up these bug fixes when they go into develop.

tbohn: OK, I see, the changes that get merged into 5.0 should ideally look like 5.0 in terms of formatting.

PR #338, opened 2016-01-14
Bugfix/maxstring

bartnijssen:
Replace hardcoded string spec with `MAXSTRING`, see issue #311


bartnijssen:
@jhamman : Let me know if I need to quash the intermediate commits or if this is OK as is (only a few lines actually changed)

jhamman:
yes- rebasing this would be nice.

Issue #335, opened 2016-01-05
Water balance errors in 5.0 classic


tbohn:
I'm getting large cumulative water balance errors in 5.0 classic, in the development branch pre-sync/4.2.c.  I haven't finished investigating - there could be erroneous settings in my global parameter file.

Some important details: I'm using a veg_library file that contains vegcover values.  The veg_parameter file also contains LAI, vegcover, and albedo climatologies for each grid cell.  I am not reading in veg_hist forcing files however.

I will try some further tests with different global parameter file settings, to see if I can isolate the problem.

tbohn:
cc-ing @jhamman and @bartnijssen

jhamman:
@tbohn - I think I've seen this before, when debugging a few other issues.  I thought I had already raised an issue here but I can't find it now (so I probably didn't).  I never ran it down far enough to give much indication of what is going wrong.  Let us know if you find anything.

tbohn:
OK, preliminary diagnosis is good: using a quick and dirty perl script to compute wb errors from the output files (I'm out of town right now and logging in via a telnet session) yields nearly 0 cumulative error for both VIC 4.2 and VIC 5.0 (3 different branches).  So it appears that the errors are simply the result of a bug in calc_water_balance_error() or put_data() rather than actual errors in fluxes or storages.  Which makes sense since we didn't really make major changes to the physics.

I'll post more when I know more.

jhamman:
I bet the bug is in `put_data`, nothing really has changed in `calc_water_balance_error`.  There were some units changes in `put_data` so that could be it.

tbohn:
After more testing, I realize that my earlier crude water balance check ran on the wrong files.  Re-running shows that, in fact, 5.0 does have water balance errors.  The errors are always equal to -evap.  Comparing results with those of 4.2, it appears that in 5.0, evap is not being subtracted from soil moisture.

Again, this is true of the develop branch pre-sync/4.2.c as well as post-sync/4.2.c.  I will focus on things like vic_run.c, surface_fluxes.c, runoff.c, etc. and will update you here when I have more information.

bartnijssen:
That seems odd, since we did not mess with the physics. Is this a bug fix that was made in 4.2 that has not made it into 5/develop yet?

tbohn:
No, there were no such water balance errors in 4.2 that needed any fixing...  Joe mentioned that he had seen water balance errors in 5.0 before but hadn't tracked them down.

[...]

PR #341, opened 2016-01-15
Fix wb errs

tbohn:
This addresses issue #335.


jhamman:
Thanks @tbohn, these fixes look great.  Are we down to zeros for the WB errors now? Regardless, these are steps in the right direction.

tbohn:
Yes, wb errors are now 0 (see comments in issue #335).

jhamman:
Well that is just grand.  I suppose I owe you a :beer:.  

be96a1d introduced a compile time bug in all drivers, once that is sorted out, this can be merged.

PR #342, opened 2016-01-15
Update global param sample

tbohn:
Addresses issue #336.

PR #347, opened 2016-01-17
Feature/fix pet

### LINKED ISSUE #276: Make PET play nice with the rest of the code
Currently, vic computes the 6 types of potential evapotranspiration (PET) by creating dummy veg tiles.  These dummy tiles cause code maintenance headaches and should be removed.  PET needs to be computed in a much less invasive manner.

### LINKED ISSUE #346: Wrong aerodynamic resistance values in 5.0
This is due to a sloppy implementation of the sync/4.2.c (I am to blame).  The problem is in `surface_fluxes.c`, in which I created a new array `iter_aero_resist_veg[]` and attempted to copy the values of `iter_aero_resist_used[]` into it.  The problem is that `iter_aero_resist_veg[]` has 3 elements but `iter_aero_resist_used[]` only has 2 elements.  This led to overstepping the array bounds and getting weird values.

So, I propose to fix this.  It will require only minor modifications to `surface_fluxes.c`.

tbohn:
This addresses issues #346 and #276.  This also contains the contents of other pull requests, unfortunately, because those other changes were necessary to generate model outputs that were sufficiently correct for me to evaluate the correctness of the PET values.

So, before merging this pull request, the other pull requests must be merged in the following order:

1. #341 (Fix wb errs)
2. #344 (Fix forceskip)
3. #342 (Update global.param.sample)

Also, at the moment the PET calculation computes PET using Penman-Monteith with the canopy resistance equal to rsmin/LAI, in other words, what the current vegetation would transpire in the absence of any limitation due to soil moisture, vapor pressure deficit, temperature, or insolation.  I'm still waiting to hear feedback from @bartnijssen in particular as to whether this is the preferred equation to use.  Of course, we can merge this as it is now and change the equation later if desired.

Also, currently, pot_evap is a scalar element of the cell_data struct, i.e. it is always allocated for each veg tile and elevation band (but at least it's not the 6-element array that it used to be).  @bartnijssen had expressed a desire to minimize memory usage and I'd suggested conditionally allocating pot_evap.  I can still do that as a minor update to this PR.  Comments are welcome...

closes #276
closes #346


tbohn:
Also, a couple files got changed when I ran uncrustify despite my not having modified the code.  Not sure why.

jhamman:
@tbohn - All the blocking PRs have been accepted.  I'm not exactly sure how you were managing your branches but I think this could use a `rebase` against `UW-Hydro/develop`.  After you run the `rebase`, @bartnijssen and I will review the PET changes.

tbohn:
OK, so I did a rebase via:
1. git checkout develop
2. git pull upstream develop
3. git checkout feature/fix_pet
4. git rebase develop feature/fix_pet
The rebase gave me the following messages:

First, rewinding head to replay your work on top of it...
Applying: Fix to PET.
Using index info to reconstruct a base tree...
M       vic/drivers/shared/include/vic_driver_shared.h
M       vic/drivers/shared/src/make_dmy.c
M       vic/vic_run/src/CalcBlowingSnow.c
M       vic/vic_run/src/lakes.eb.c
M       vic/vic_run/src/photosynth.c
M       vic/vic_run/src/root_brent.c
M       vic/vic_run/src/surface_fluxes.c
Falling back to patching base and 3-way merge...
Auto-merging vic/vic_run/src/surface_fluxes.c
Auto-merging vic/drivers/shared/include/vic_driver_shared.h

So, then I attempted to push this to origin, but got the following message:

 ! [rejected]        feature/fix_pet -> feature/fix_pet (non-fast-forward)
error: failed to push some refs to 'https://github.com/tbohn/VIC.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

So, did I do the rebase wrong?  The code looks the same as before I did all this, as far as I can tell.  Should I do a git pull from my origin version of feature/fix_pet, or will this only replace the rebase with the previous state (whatever that was)?

tbohn:
So then I tried a git pull origin feature/fix_pet, followed by the same rebase as before.  Still can't push it to origin due to its being behind the origin.

I'm going to try clicking on the "update branch" button here to see if that helps.

tbohn:
It seems like rebasing always results in my local branch being behind the origin branch.

jhamman:
pushing after a rebase usually requires a "force push".  `git push -f origin feature/fix_pet`

tbohn:
OK, I did the push -f.  Actually, even before I did that, I took a look at the updated diff and it appeared to only contain changes associated with fixing PET.  So perhaps it's ready to be reviewed?

vic/vic_run/src/compute_pot_evap.c
```diff
@@ -27,120 +27,60 @@
...
+    net_rad = net_short + net_longwave;
+    if (!overstory) {
+        ra_veg = aero_resist[0];
+    }
+    else {
+        ra_veg = aero_resist[1];
     }
+    Epot_veg =
+        penman(tair, elevation, net_rad, vpd, ra_veg, rc, rarc) *
+        dt / SEC_PER_DAY;
```
jhamman: is `dt / SEC_PER_DAY` equal to `STEPS_PER_DAY`?

Same comment for line 83.

vic/vic_run/src/surface_fluxes.c
```diff
@@ -602,6 +593,11 @@ surface_fluxes(bool                 overstory,
                     INCLUDE_SNOW = false;
                 }
 
+                if (iter_snow.snow) {
+                    iter_aero_resist_veg[0] = iter_aero_resist_used[0];
+                    iter_aero_resist_veg[1] = iter_aero_resist_used[1];
+                }
```
jhamman: why is this change needed?

vic/vic_run/src/surface_fluxes.c
```diff
@@ -797,49 +793,17 @@ surface_fluxes(bool                 overstory,
...
-                step_aero_resist[p][1] = aero_resist[p][1] *
-                                         stability_factor[1];
-            }
-        }
-
-        // Finally, compute pot_evap
-        compute_pot_evap(veg_class, dmy, gp->dt, atmos->shortwave[hidx],
-                         iter_soil_energy.NetLongAtmos, Tair, VPDcanopy,
-                         soil_con->elevation, step_aero_resist, iter_pot_evap);
+        compute_pot_evap(gp->dt, vic_run_veg_lib[veg_class].rmin,
```
jhamman: I don't think you need to pass in `gp->dt` since that is available as an `extern`.  It is also better to use `STEPS_PER_DAY`.

jhamman:
@tbohn - nice work on the rebase.

I had a quick look.  Love the 400 deletions to just 200 new lines.  This makes the VIC PET code much more tractable. I'll let you and @bartnijssen sort out which resistance to use.

vic/vic_run/src/surface_fluxes.c
tbohn: Because in the case of snow, aerodynamic resistance is modified in snow_intercept().  We need the "final" aerodynamic resistance over vegetation to be used in the weighted average in func_surf_energy_bal().

Now that I think of it, I wonder if I shouldn't just use the "veg" version for everything in the case of snow.  Perhaps I should add a case to func_surf_energy_bal() to cover this.  Either way, this updated value needs to be piped to func_surf_energy_bal() as I'm doing here.

vic/vic_run/src/compute_pot_evap.c
tbohn: OK, done.

vic/vic_run/src/surface_fluxes.c
tbohn: OK, done (got rid of gp->dt in argument list).

tbohn:
As per conversation with @bartnijssen and @jhamman, am going with Penman-Monteith equation (using the current architectural resistance and LAI), since that is both a diagnostic variable of VIC's actual ET and also something that a user is much less likely to compute on their own (than, say, the simple Penman equation for some reference crop).

Also per that conversation, will keep it as a scalar rather than conditionally allocating it.

jhamman:
Looks good @tbohn.  Two remaining items: 

1.  Add a brief description of these changes to the docs here: http://vic.readthedocs.org/en/develop/Development/ReleaseNotes/#backwards-incompatible-changes
2.  Do a rebase/squash on these changes so this PR is just one commit.  The basic idea is that log messages like "Updated another comment" aren't useful in the future.  So, we just squash them all into one commit.  This is a decent ref on how to do that: http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git?answertab=votes#tab-top

tbohn:
I tried doing a squash via 'git reset -soft HEAD~9' followed by a commit.  But when I tried to push this to origin, it complained that now the tip of my branch is behind origin by several commits.  I'll try to see if I can fix this.

tbohn:
Can't seem to make the squash work.  Is this a show-stopper?

tbohn:
Dangit, I've lost all my changes, both locally and in my fork, due to trying various types of rebasing to get around the errors and inability to push to origin (ending with a push --force, which I will make sure to be much more cautious about in the future).  Fortunately I have a back-up copy that's only a little out of date - hopefully I can reconstruct all the changes again.  I'll have to submit a new pull request.

PR #348, opened 2016-01-18
Cleanup runoff.c

tbohn:
This addresses issue #345.

This contains code changes from PR #347 (Feature/fix pet).

PR #349, opened 2016-01-19
Clean up runoff.c

tbohn:
This addresses issue #348.  Note that this replaces my earlier pull request on the same topic, as my earlier pull request was done from a branch built on top of the feature/fix_pet branch, and I was having trouble separating the changes of the two PRs.  In this new PR, the branch is directly off of develop.

closes #348


vic/vic_run/src/runoff.c
```diff
@@ -48,7 +48,6 @@ runoff(cell_data_struct  *cell,
     size_t                     time_step;
     int                        last_index;
     int                        tmplayer;
-    size_t                     frost_area;
     int                        fidx;
```
jhamman: `fidx` should be a `size_t`, that will take care of these warnings: https://travis-ci.org/UW-Hydro/VIC/jobs/103274507#L171-L183

tbohn: My current solution to this is to keep fidx as an int so that the two loops that decrement it can evaluate whether fidx >= 0 (for fidx as size_t, decrementing fidx beyond 0 is meaningless and the loop condition of fidx >= 0 will always be true, at least according to the warning statement).  To avoid warnings on comparing objects of different types, I cast options.Nfrost as int in the 6 or so instances where fidx is involved in an expression with options.Nfrost.

The other solution I looked into was making fidx a size_t but modifying the decrementing loops.  One modification involved adding 1 to the loop start/end values (i.e. fidx = options.Nfrost instead of options.Nfrost - 1; and fidx > 0 instead of fidx >= 0) and subtracting 1 from fidx inside the arrays (e.g. evap[][fidx-1]) but looking at it I felt this would make the code harder to understand / less intuitive.

Hopefully casting options.Nfrost as int is acceptable?

PR #357, opened 2016-01-21
Consistent default VEGCOVER behavior across modes.

### LINKED ISSUE #337: Unable to *not* supply vegcover values in both veg_lib and veg_param simultaneously
Not sure whether this is intentional or a bug.

At present, in VIC 5.0, we can't supply VIC with both veg_lib and veg_param files that don't contain vegcover values.  This is because the global option VEGCOVER_SRC can only be set to one of: FROM_VEGLIB or FROM_VEGPARAM.  If it's set to FROM_VEGLIB but the veg_lib file doesn't contain vegcover values (and VEGLIB_VEGCOVER accurately reflects this by being FALSE), VIC will exit with an error.

So, my question is: do we want to require users to supply vegcover values in the veglib?  They'll need new veglib files.  This is easy enough to do - it's a minor modification to the old NLDAS veg lib to add 12 vegcover values of 1.0 to each veg class, and that should theoretically allow users to reproduce old behavior.  If this is the way we want to go, then we should post a new version of the NLDAS veg lib for VIC 5.0 that contains these values.

Personally I'm fine with this solution and would prefer it.  But @bartnijssen and @jhamman  I need to know your opinions on this before I can propose a specific solution.

If we require vegcover values, I'd be happy to create a new NLDAS veg lib for the "samples" directory.

If we want to allow the users to not have anything to do with vegcover values, then we need to modify the logic in VIC that doesn't allow vegcover to be omitted from both veg_lib and veg_param files.  Could be a simple matter of initializing VEGCOVER_SRC to MISSING and then adding a case for MISSING that would set vegcover to 1.0 everywhere/all time.

tbohn:
Created new values for veg parameter *_SRC options:
- FROM_DEFAULT - primarily intended for VEGCOVER, to indicate setting to 1.0
- FROM_VEGHIST - indicates that a veg parameter should take its value from the veg_hist forcing file(s)

VEGCOVER_SRC is now initialized to FROM_DEFAULT.

When VEGCOVER_SRC == FROM_DEFAULT, all vegcover values are set to 1.0.

When any of LAI_SRC, VEGCOVER_SRC, or ALBEDO_SRC == FROM_VEGHIST, their values are taken from the veg_hist forcing file(s).  This is not yet implemented in image mode.

Values of FROM_VEGLIB and FROM_VEGPARAM have the same behavior as before for classic mode (take values from the veg_lib or veg_parameter files, respectively).  For image mode, both of these values indicate to take values from the veg_parameter file, since all veg-related parameters are read from the veg_parameter file in image mode.  This can be modified later if desired.

Additionally, this commit removes all instances of "LAI_FROM_VEGLIB" and "LAI_FROM_VEGPARAM", as these were only present for backwards compatibility.

Similarly, this commit removes the logic in get_global_param() for handling "GLOBAL_LAI", as this also is only present for backwards compatibility.

closes #337


tbohn:
A little more explanation: the purpose of this commit is primarily to allow classic mode to turn off the vegcover feature.  Previously the default was to read vegcover from the veg library, which required the user to put vegcover values in their library.  Now, VIC can once again read a completely traditional set of input files that have no mention of vegcover anywhere.

In addition, this allows the user to more easily turn vegcover on/off and select the source of these values.  Previously, if the user specified veg_hist forcing files in the global parameter file, VIC would use them for LAI, VEGCOVER, and ALBEDO; the only way to turn that behavior off would be to comment out the lines in the global parameter file describing the veg_hist forcing files.  Now, the user only needs to change LAI_SRC, VEGCOVER_SRC, and ALBEDO_SRC to the appropriate "FROM_".  Furthermore, the user can now get different variables from different sources, e.g., LAI_SRC can = FROM_VEGPARAM while VEGCOVER_SRC can = FROM_VEGHIST, etc.

vic/drivers/image/src/get_global_param.c
```diff
@@ -621,40 +621,70 @@ get_global_param(FILE *gp)
...
+                }
+                else if (strcasecmp("FROM_VEGPARAM", flgstr) == 0) {
+                    options.VEGCOVER_SRC = FROM_VEGPARAM;
+                }
+                else if (strcasecmp("FROM_VEGLIB", flgstr) == 0) {
+                    options.VEGCOVER_SRC = FROM_VEGLIB;
                 }
                 else {
-                    options.LAI_SRC = LAI_FROM_VEGLIB;
+                    options.VEGCOVER_SRC = FROM_DEFAULT;
```
jhamman: It would be better to fail if the `VEGCOVER_SRC` option is not recognized.  The default should be `FROM_DEFAULT` as you have it but if someone has this line in their param file, they will get the default value, which is not what they wanted:  
```
VEGCOVER_SRC LAI_FROM_VEGPARMA   # typo!
```

So you just need to put one more `else if` statment in there for `FROM_DEFAULT`, then fail if you get to the else.

vic/drivers/image/src/get_global_param.c
```diff
@@ -621,40 +621,70 @@ get_global_param(FILE *gp)
...
+            else if (strcasecmp("LAI_SRC", optstr) == 0) {
+                sscanf(cmdstr, "%*s %s", flgstr);
+                if (strcasecmp("FROM_VEGHIST", flgstr) == 0) {
+                    options.LAI_SRC = FROM_VEGHIST;
+                }
+                else if (strcasecmp("FROM_VEGPARAM", flgstr) == 0) {
+                    options.LAI_SRC = FROM_VEGPARAM;
+                }
+                else {
+                    options.LAI_SRC = FROM_VEGLIB;
```
jhamman: possible same comment as https://github.com/UW-Hydro/VIC/pull/357/files#r50583033

vic/drivers/image/src/get_global_param.c
```diff
@@ -621,40 +621,70 @@ get_global_param(FILE *gp)
...
+            else if (strcasecmp("ALB_SRC", optstr) == 0) {
+                sscanf(cmdstr, "%*s %s", flgstr);
+                if (strcasecmp("FROM_VEGHIST", flgstr) == 0) {
+                    options.ALB_SRC = FROM_VEGHIST;
+                }
+                else if (strcasecmp("FROM_VEGPARAM", flgstr) == 0) {
+                    options.ALB_SRC = FROM_VEGPARAM;
+                }
+                else {
+                    options.ALB_SRC = FROM_VEGLIB;
```
jhamman: possible same comment as https://github.com/UW-Hydro/VIC/pull/357/files#r50583033

vic/drivers/image/src/get_global_param.c
```diff
@@ -1140,21 +1170,6 @@ get_global_param(FILE *gp)
...
-                "contains 1 line of 12 monthly LAI values for EACH veg "
-                "tile in EACH grid cell, and you MUST specify "
-                "\"VEGPARAM_LAI\" as \"TRUE\" in the global parameter "
-                "file.  Alternatively, if you want VIC to read LAI values "
-                "from the veg library file, set \"LAI_SRC\" to "
-                "\"LAI_FROM_VEGLIB\" in the global parameter file.  In "
-                "either case, the setting of \"VEGPARAM_LAI\" must be "
-                "consistent with the contents of the veg param file (i.e. "
-                "whether or not it contains LAI values).");
-    }
```
jhamman: Do we need any additional checks that all these new options work together?  This would be the place.

jhamman:
This looks pretty good @tbohn.  I just had a few comments on the handling of global parameter options.

vic/drivers/image/src/get_global_param.c
tbohn: Actually, I think we should remove some more of these checks in this particular file, since netcdf i/o functions called by vic_init() and vic_force() will handle attempts to read variables that don't exist.

However, your point *does* apply to the classic driver.  I agree, I will see if there are any other cases to handle.

tbohn: Further clarification: in image mode, options.VEGPARAM_LAI and options.VEGLIB_LAI don't really make sense for the reason I mentioned above.

jhamman: Sounds good. My comment can apply to the classic driver then.

jhamman:
@tbohn - can you pull in the latest develop branch here. The travis tests should pass now.

tbohn:
The image mode build failed, but all it says is that the node doing the compiling didn't return a status after 10 minutes, so we don't know what the error message was...

jhamman:
I re-triggered the build.  I think there is something going wrong with some of the travis nodes tonight because I've had this happen a few times recently.

tbohn:
Hey... it compiled!

jhamman:
Yep - this looks mostly done, after you resolve the merge conflicts, I'll take one last look.

Joe

vic/drivers/image/src/get_global_param.c
tbohn: Yes, just to confirm, I added checks on options.*_SRC == FROM_VEGHIST, requiring that the variable in question have actually been supplied as a forcing.  Again, this is just for classic mode, as netcdf i/o will take care of that in image mode.

tbohn:
This is strange.  I see the message about conflicts that must be resolved, but I can't seem to find those conflicts.  On the command line, I did a git pull from origin to bring the merge back to my workspace, but I don't see any merge conflicts there.  I wish the message here on GitHub had a link to a list of those conflicts...  Am I missing something?

Perhaps I should have done the merge on the command line and then, after resolving the conflicts, pushed to origin...

jhamman:
You want to merge in `upstream/develop` (e.g. `UW-Hydro/develop`).  Most likely your merge was attempting to reconcile conflicts with `origin/develop` (e.g. `tbohn/develop`).  Try `git pull upstream develop`.

tbohn:
OK, I resolved the conflicts in samples/global.param.sample.classic.txt.

tbohn:
@jhamman this is ready to go as far as I can tell.

jhamman:
Yep.  Looks good to me.  Merging now.

tbohn:
Thanks @jhamman !

PR #359, opened 2016-01-22
Move veg_hist logic to driver.

### LINKED ISSUE #275: Move veg_hist logic outside of vic_run()
Time-varying veg parameters such as LAI, vegcover, and albedo should all be passed to vic_run() via a single data structure, containing only the elements of these arrays that apply to the current time step.  The logic to choose the source of these parameter values (from veg_lib, veg_parameter file, or veg_hist file) should happen outside of vic_run().  vic_run() should not need to know anything about the current record (or month).

### LINKED ISSUE #325: Enable reading of veg_hist from netcdf files
Need to add this capability to the image driver.  @bartnijssen and @jhamman I might want to discuss the structure of these files with you before I get started.

I suppose that the netcdf file would require a uniform number of "levels" in the veg class dimension across all grid cells.  Veg classes that aren't present would just have a missing/fill value.  Any other ideas?

It might also be worth revisiting how it's handled in ascii.  The way I've implemented it so far is to have a column for each veg tile.  This results in different numbers of columns per cell, and necessitates sending veg param information to the reading functions.  But it certainly saves space...  I'm not sure how else to do this other than having a column for every veg class in every file, and just use a missing value (nodata) in those columns for veg classes that aren't present.  This would simplify the VIC code but dramatically increase the size of the veg_hist files.  Any other ideas are welcome...

tbohn:
Moved veg_hist logic out of vic_run().

Created new function update_step_vars() which currently only updates the veg_var
 structure to store the current veg_hist variables.  This is called from vic_classic() in classic mode and from vic_image_run() in image mode.

Added logic to image mode vic_force() to read timeseries (non-climatological) of LAI, albedo, and vegcover from the 2nd forcing file, conditional on those variables' options.VARNAME_SRC being set to FROM_VEGHIST.

In both classic and image mode, the veg_hist array is initialized with climatological values from veg_con, and then optionally overwritted with values from the veg_hist file if options.VARNAME_SRC is set to FROM_VEGHIST.

closes #275 
closes #325


vic/drivers/shared/src/update_step_vars.c
```diff
@@ -0,0 +1,68 @@
...
+    veg_var = all_vars->veg_var;
+
+    Nbands = options.SNOW_BAND;
+
+    /* Set number of vegetation types */
+    Nveg = veg_con[0].vegetat_type_num;
+
+    /* Assign current veg characteristics */
+    for (iveg = 0; iveg < Nveg; iveg++) {
+        veg_class = veg_con[iveg].veg_class;
```
jhamman: This variable is set but not used.

vic/drivers/image/src/vic_force.c
```diff
@@ -192,6 +194,105 @@ vic_force(void)
...
+                    veg_hist[i][vidx].albedo[j] =
+                        veg_con[i][vidx].albedo[dmy[current].month - 1];
+                    veg_hist[i][vidx].LAI[j] =
+                        veg_con[i][vidx].LAI[dmy[current].month - 1];
+                    veg_hist[i][vidx].vegcover[j] =
+                        veg_con[i][vidx].vegcover[dmy[current].month - 1];
+                }
+            }
+        }
+    }
```
jhamman: Does this need to be done here?  This seems like it will be done elsewhere and, as it is now, results in overwriting of the `veg_hist` values even if none of the variables below `== FROM_VEGHIST`.

vic/drivers/image/src/vic_force.c
```diff
@@ -220,6 +321,17 @@ vic_force(void)
...
+        for (v = 0; v < options.NVEGTYPES; v++) {
+            vidx = veg_con_map[i].vidx[v];
+            if (vidx != -1) {
+                for (j = 0; j < NF; j++) {
+                    if (veg_hist[i][vidx].vegcover[j] < MIN_VEGCOVER) {
+                        veg_hist[i][vidx].vegcover[j] = MIN_VEGCOVER;
+                    }
+                }
+            }
+        }
```
jhamman: This isn't so much a validation as it is a correction.  Maybe we should print a warning or raise an error here?

jhamman:
This looks good too @tbohn.  A few comments on the implementation, but otherwise this is a nice cleanup of the veghist code.  

This is blocked by #357

vic/drivers/shared/src/update_step_vars.c
tbohn: OK, fixed.

vic/drivers/image/src/vic_force.c
tbohn: This overwriting is intentional.  My primary reason was ease of coding (laziness).  Here, in vic_force(), the loop over months of the veg_con climatology can be done easily, and then veg_hist has the climatological values, already "disaggregated" into snow steps, and carries these out to the rest of the code.  If I were to delay this assignment until update_step_vars(), when we are choosing whether to read the values from netcdf files (i.e., no overwrite, just a single write that chooses between veg_con and netcdf), then update_step_vars() would have to know the current month.

So, when we exit vic_force(), the veg_hist arrays are *the* place to look for the LAI etc values, regardless of which option was chosen.

These all result in very simple code in update_step_vars().

This won't affect current performance if you aren't supplying a veg_hist file - this assignment is the same assignment that existed prior to my edits.  If you're reading from a veg_hist file, it does result in a second assignment.  Do you think that that will have a noticeable effect on performance in that case?

vic/drivers/image/src/vic_force.c
tbohn: You're right, it's a correction.  Sure, we could raise a warning.

The reason for this, by the way, is because we divide by LAI and albedo by vegcover and even if vegcover > 0, it could be so extremely small as to lead to large precision errors (although now that I think of it, I should make sure that this isn't done for tiles that are explicitly labeled as bare soil - I've run this code in regions with pure bare soil tiles, but what likely happened is that the bare soil tile was divided into a "vegetated" portion with LAI of 0 and a "bare" portion also with LAI of 0, resulting in the same end result but with twice the work... I can try to add a case to exclude this for pure bare soil tiles).

tbohn: Can we raise a warning and still make the correction?  Or is the goal to be strict about inputs?

jhamman:
@tbohn - can you pull in the latest develop branch here.  The travis tests should pass now.

tbohn:
@jhamman OK, I've pulled in the latest develop branch (via merging from create_default_no_soil_evap branch, so that it has the resolution of the merge conflicts that I did there).

I'm still looking for clarification: is it OK to perform the correction of vegcover relative to MIN_VEGCOVER, with a warning message, or did you mean that the warning message would take the place of the correction?  I'm guessing that the warning would be in addition to the correction to avoid division by 0.  I'll go ahead and implement that but if that's not what you meant, please let me know.

tbohn:
@jhamman this is ready for your review...

vic/drivers/image/src/vic_force.c
bartnijssen: I think any time that we change a user defined input - even if there are good physical reasons to do so - we should not do so silently, because users should be aware that their inputs are modified. For example, if they are analyzing their results assuming a certain LAI value, then they need to be aware that this has been changed.

It may actually be nice to put all warning messages in the documentation where we can include a more complete explanation of why the warning is necessary (for example the explanation from @tbohn above). That way we can refer in the warning to the correct place in the documentation and keep the warning short. 

So as far as I am concerned as long as we warn I think this is fine.

tbohn: So, I did add code to raise a warning, and pushed it to origin at least 1 day ago, but it's not showing up here...

jhamman: Okay, for this PR, lets just issue a warning.  I'll open a separate issue for connecting warning and error messages to the docs.

vic/drivers/image/src/vic_force.c
jhamman: I don't care about the performance.  I'm mostly concerned with unintended outcomes.  It sounds like this is the right way to go.

vic/drivers/image/src/vic_force.c
tbohn: Let me rephrase that - the new code shows up on the "files changed" tab, but for some reason the "conversation" tab hasn't been updated - the code is out of date.

jhamman:
This looks good.  Merging now.

vic/drivers/image/src/vic_force.c
jhamman: Yes, it is because the line I commented on (337) hasn't changed, therefor the diff didn't change.  If I had highlighted one of the lines that you changed, the diff would have changed (actually it would be collapsed and labeled as outdated).

PR #363, opened 2016-01-23
Split global.param.sample into classic and image versions.

### LINKED ISSUE #356: Some sample global parameter files contain outdated options
Global parameter files are found in the following locations:
```
samples/global.param.sample
tests/examples/global*
tests/release/global*
tests/system/global*
```
Many of these contain outdated options like `MEASURE_H`, `STARTHOUR`, `TIME_STEP`, `PLAPSE`, etc. that have been removed in 5.0.  We need to update these to work for 5.0.

tbohn:
closes #356


tbohn:
This PR accomplishes the split.  I imagine that further changes are necessary.

Note that I removed VEG_LIB but kept VEG_PARAM.  In other words, I'm calling the netcdf file that contains veg information the VEG_PARAM file, not the VEG_LIB file.  If you'd rather have it the other way, let me know and I can switch it.

jhamman:
nit: can we rename the files so they have a suffix that indicates they are text files (because they are).  `.txt` is fine by me.

Otherwise, this looks good.

tbohn:
OK, renamed.

PR #388, opened 2016-02-10
Added support for reading lake parameters in image driver

### LINKED ISSUE #326: Implement reading of lake parameters in 5.0 image mode
Currently `vic_init()` contains just a placeholder for reading lake parameters.

### LINKED ISSUE #382: better name for vegcover?
The veg parameter "vegcover" denotes the area fraction of a given veg tile that is covered by plants (as opposed to exposed soil between the plants).  But "vegcover" sounds similar to the definition of Cv (the area fraction of a veg tile) and the exposed soil sounds similar to a bare soil tile, leading to some confusion.  My attempts to make the difference clear in the comments describing the variables result in long descriptions.

I'm wondering if it might help to change the name of this variable to something like "fcan" (= canopy coverage fraction).  The exposed soil between plants could be viewed as gaps in the canopy.  My one reservation about this term is that users may associate a canopy more with forest (i.e., overstory) and less with low-lying vegetation.  Any opinions?

tbohn:
This PR added support for reading lake parameters to the image driver.  Because I used the new name "fcanopy" instead of the old name "vegcover" when using tonic to generating netcdf files for testing, I also changed that variable's name to "fcanopy" in VIC.

Thus, we have:
closes #326 
closes #382


vic/drivers/classic/src/get_global_param.c
```diff
@@ -1289,46 +1297,46 @@ get_global_param(FILE *gp)
...
-                "from the veg param file, set \"VEGCOVER_SRC\" to "
+                "\"VEGLIB_FCAN\" as \"TRUE\" in the global parameter "
+                "file.  Alternatively, if you want VIC to read fcanopy values "
+                "from the veg param file, set \"FCAN_SRC\" to "
                 "\"FROM_VEGPARAM\" in the global parameter file.  In "
-                "either case, the setting of \"VEGLIB_VEGCOVER\" must be "
+                "either case, the setting of \"VEGLIB_FCAN\" must be "
                 "consistent with the contents of the veg lib file (i.e. "
-                "whether or not it contains vegcover values).");
+                "whether or not it contains fcanopy values).");
```
jhamman: add `VEGCOVER_SRC` to the list of deprecated variables in this file (not right here but somewhere we have a long list of variables that were in 4.2 but not in 5.0.

vic/drivers/image/src/vic_init.c
```diff
@@ -1101,15 +1109,253 @@ vic_init(void)
...
+                                "fraction %f\n", i, j, lake_con[i].Cl[j], j - 1,
+                                lake_con[i].Cl[j - 1]);
+                    }
+                }
+            }
+        }
+
+        // compute other lake parameters here
+        for (i = 0; i < local_domain.ncells_active; i++) {
+            compute_cell_area(&(soil_con[i]));
```
jhamman: area should come from the domain file for the image driver.

vic/drivers/shared/include/vic_driver_shared.h
```diff
@@ -445,6 +445,8 @@ void collect_eb_terms(energy_bal_struct, snow_data_struct, cell_data_struct,
 void collect_wb_terms(cell_data_struct, veg_var_struct, snow_data_struct,
                       double, double, double, int, double, int, double *,
                       double *, out_data_struct *);
+void compute_cell_area(soil_con_struct *);
```
jhamman: Can we put this in the classic driver.  I don't think we need/want it in the image driver(s).

vic/vic_run/src/lakes.eb.c
```diff
@@ -26,6 +26,7 @@
 
 #include <vic_def.h>
 #include <vic_run.h>
+#include <vic_driver_shared.h>
```
jhamman: I think we've been trying to keep the driver headers out of `vic_run`.  What is this needed for?

vic/drivers/shared/include/vic_driver_shared.h
tbohn: OK, will do.

vic/drivers/image/src/vic_init.c
tbohn: OK, will do.

vic/drivers/classic/src/get_global_param.c
tbohn: OK, will do.

vic/vic_run/src/lakes.eb.c
tbohn: This is because the get_depth(), get_volume() etc. functions are stored in a file that's part of the driver code.  I can break them out into separate files and move them to vic_run/.

jhamman: sounds like a good plan.

vic/drivers/image/src/vic_init.c
tbohn: Question: is it sufficient for now to just set soil_con[i].cell_area = global_domain.locations[i].area?

The classic driver doesn't as yet have a global_domain structure, right?  If it did, we could get rid of soil_con.cell_area across the board.  Either way, it seems like this would be a separate isue...

vic/vic_run/src/lakes.eb.c
tbohn: Actually, this is because the lake module needs the initialize_lake() function (which is currently in the driver section).  When a lake shrinks to nothing, and then re-appears, the state of the new lake needs to be initialized - that's why the lake module needs the function.

We could move initialize_lake() into vic_run/ if that would help.  vic_init() is already including vic_run.h (which is where we'd put the definition of initialize_lake()).

jhamman: that's fine too.  As long as we don't have any driver code sitting in `vic_run`.

tbohn: So, you don't consider initialize_lake() to be driver code?  That makes
sense to me (not considering it to be driver code).  However, the other
initialize_* functions are still sitting in the drivers/shared/ section.

I guess for now we can consider initialize_lake() to be "promoted" to
vic_run/ due to its being used by the lake module.

jhamman: If its required by `vic_run`, its not driver code. Maybe the drivers should handle the creation/initialization of lakes but I think that is a separate issue.

jhamman:
@tbohn - once you resolve these merge conflicts, I'll give this one more read through.  From what I can tell, it looks good to go.

tbohn:
OK, I rebased to upstream/develop.  However, then to push to origin, I had to push --force.  Hopefully that doesn't mess up the history in this PR...

vic/drivers/classic/src/compute_cell_area.c
```diff
@@ -0,0 +1,72 @@
...
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License along with
+* this program; if not, write to the Free Software Foundation, Inc.,
+* 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+******************************************************************************/
+
+#include <vic_def.h>
+#include <vic_run.h>
+#include <vic_driver_shared.h>
```
jhamman: `<vic_driver_classic.h>`?

tbohn: OK, fixed.

tbohn:
So, I've addressed all of your comments...  Can this be merged?

vic/drivers/classic/src/get_global_param.c
jhamman: Did this get done?  I don't see anywhere in these code changes where you raise an error for using `VEGPARAM_VEGCOVER` or `VEGLIB_VEGCOVER`.

tbohn: It got done, but I didn't add the validation/error for the old values.  I'll add that.

jhamman:
@tbohn - just that one last comment then we're ready to merge.  Let me know once you've pushed again and the travis build is green.

vic/drivers/classic/src/get_global_param.c
tbohn: So, actually, not only did I not add such an error, but I removed the old error for using "GLOBAL_LAI" because I figured that option is old enough by now that we can stop maintaining that error.  So, I guess my question now is: is there a limit to how old deprecated options can be before we stop checking for them and raising errors?  Should I reinstate the check on GLOBAL_LAI?

jhamman: I think options that were available in 4.1.2 or 4.2 should still be explicitly deprecated.  When was `GLOBAL_LAI` removed?

tbohn: GLOBAL_LAI was removed starting in 4.1.2, so I guess it's OK to get rid of the check on that one.

tbohn:
OK, pushed and built.

jhamman:
Okay.  Thanks @tbohn.  Merging now...

PR #389, opened 2016-02-11
Added support for reading blowing snow parameters in image driver

tbohn:
This closes #327.

This is a branch off of PR #388.  Thus at present it contains the commits from that PR.  But as soon as that PR is taken, the diff for this PR should collapse down to a single file: vic/drivers/image/src/vic_init.c.

PR #391, opened 2016-02-13
Added support for reading carbon parameters in image driver

### LINKED ISSUE #390: Implement reading of carbon parameters in 5.0 image driver
Carbon parameters from the veg_lib file are not currently read in.

tbohn:
This is a branch off of PR #389.

Closes #390.


vic/drivers/classic/src/read_veglib.c
```diff
@@ -163,10 +163,10 @@ read_veglib(FILE   *veglib,
             /* Carbon-cycling parameters */
             if (options.VEGLIB_PHOTO) {
                 fscanf(veglib, "%s", tmpstr); /* photosynthetic pathway */
-                if (!strcmp(tmpstr, "C3")) {
+                if (!strcmp(tmpstr, "0") || !strcmp(tmpstr, "C3")) {
                     temp[i].Ctype = PHOTO_C3;
                 }
-                else if (!strcmp(tmpstr, "C4")) {
+                else if (!strcmp(tmpstr, "1") || !strcmp(tmpstr, "C4")) {
```
jhamman: Since you're deprecating `0` and `1`, can you print a warning statement if those options are used?

tbohn: OK, will do.  Except that it's the strings "C3" and "C4" that I'm deprecating.  And the reason for that is that I had trouble figuring out how to get them to work correctly in tonic - I tried for hours - and finally relented and made them ints.  So, tonic can read either of them (string or int), but will convert to the 0/1 values for writing to netcdf.  So, VIC is reading them as ints in image mode.

jhamman: Why don't we figure out how to fix your `tonic` code?  If you can show me what you weren't able to get to work, I'm sure I can help you get it to work. No need to change VIC.

jhamman: okay, after further thought and an offline conversation with @tbohn - I think this is fine for now.

jhamman:
merging now...

vic/drivers/classic/src/read_veglib.c
tbohn: :-)

PR #393, opened 2016-02-26
Update write_data.c for binary output when OUTPUT_FORCE=TRUE

lizaclark:
dmy structure is NULL if OUTPUT_FORCE option is TRUE; thus, dmy->year, etc. cannot be accessed in write_data.c

### DIFF
--- a/src/write_data.c
+++ b/src/write_data.c
@@ -110,11 +110,15 @@ void write_data(out_data_file_struct *out_data_files,
     tmp_fptr = (float *)calloc(N_OUTVAR_TYPES*options.Nlayer*options.SNOW_BAND,sizeof(float));
     tmp_dptr = (double *)calloc(N_OUTVAR_TYPES*options.Nlayer*options.SNOW_BAND,sizeof(double));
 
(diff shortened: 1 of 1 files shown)


jhamman:
The question I have is why is `dmy` null.  It should have been populated when `mtclim` was being run...?

lizaclark:
`dmy` isn't passed to `write_forcing_file.c`, so `write_forcing_file.c` creates a `dummy_dmy = NULL;` to pass into `write_data.c` (the program that feeds `atmos_data_struct` variables to `out_data_struct`), presumably because VIC doesn't write the time when writing forcing output.

jhamman:
I'm fine with merging this.  Since we're pretty close to the VIC 5 release, I'd like to hold off on releasing 4.2.d until just before we release 5.0. 4.2.d will then be the basis for the VIC 4 support branch.

bartnijssen:
@jhamman : OK - so I will leave this one open for you to merge

jhamman:
Thanks @lizaclark!

PR #394, opened 2016-03-01
Fixed output file bugs in classic mode.

yixinmao:
When no OUTFILE specified in global file, output default files;
When OUTFILE is specified but no corresponding OUTVAR specified, raise
an error.

### DIFF
--- a/vic/drivers/classic/src/vic_classic.c
+++ b/vic/drivers/classic/src/vic_classic.c
@@ -67,6 +67,7 @@ main(int   argc,
     int                   cellnum;
     int                   startrec;
     int                   ErrorFlag;
+    size_t                filenum;
     dmy_struct           *dmy;
     atmos_data_struct    *atmos;
     veg_hist_struct     **veg_hist;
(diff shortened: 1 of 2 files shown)


vic/drivers/classic/src/vic_classic.c
```diff
@@ -67,6 +67,7 @@ main(int   argc,
     int                   cellnum;
     int                   startrec;
     int                   ErrorFlag;
+    unsigned int          filenum;
```
jhamman: let's make this a `size_t`.

vic/drivers/classic/src/vic_classic.c
```diff
@@ -114,6 +115,11 @@ main(int   argc,
     fclose(filep.globalparam);
     filep.globalparam = open_file(filenames.global, "r");
     parse_output_info(filep.globalparam, &out_data_files, out_data);
+    for (filenum = 0; filenum < options.Noutfiles; filenum++) {
+        if (out_data_files[filenum].nvars == 0) {
+            log_err("Must specify at least one variable for each output file. Zero found.");
```
jhamman: can we change this slightly to something like: 
```C
log_err("No output variables were set in OUTFILE %zu.  Must set at least one output variable (OUTVAR) for each OUTFILE", filenum);
```

jhamman:
Thanks, this looks great.

yixinmao:
@jhamman Updated PR as suggested.

yixinmao:
@jhamman : Ran uncrustify on the changed two files. (it seems that several other files are not under the desired format if I understood it correctly, but I didn't change them)

jhamman:
Thanks @yixinmao - this is great!

vic/drivers/classic/src/vic_classic.c
Sibada: What api vic has?


来自 魅族 MX4 Pro

PR #399, opened 2016-03-21
Commit to fix memory error in get_nc_var_text

bartnijssen:
@jhamman: I think this is the fix that is needed. This is just for review

Let me know how to squash all these "merge" commits.


jhamman:
cc @wietsefranssen - this should fix the bug you were seeing.

jhamman:
@bartnijssen - I think we should just go ahead and merge this as is. A rebase/squash isn't going to get rid of these commits since they are already (most of them at least) in the develop branch.  I'm honestly not sure why they are showing up here.

jhamman:
thanks @bartnijssen.

wietsefranssen:
Thanks! It indeed solved the error!

PR #402, opened 2016-03-21
Added state file i/o in image driver.

tbohn:
Note that I am not finished testing this.  I am trying to ensure that VIC can read the state files produced by tonic (as well as its own state files).  But at least the code differences are visible now.


jhamman:
@tbohn - I just briefly looked through this and it looks good. Thanks for putting it together. We'll give it a full code review but you'll need to resolve the merge conflict first.

bartnijssen:
@tbohn : can you resolve the merge conflicts before I review?

tbohn:
@bartnijssen I will try my best to get to this in the next couple of days, but can't do it right away.

vic/drivers/image/src/vic_restore.c
```diff
@@ -29,39 +29,987 @@
...
+    }
+
+    // snow age: snow[veg][band].last_snow
+    for (m = 0; m < options.NVEGTYPES; m++) {
+        d4start[0] = m;
+        for (k = 0; k < options.SNOW_BAND; k++) {
+            d4start[1] = k;
+            get_scatter_nc_field_int(filenames.init_state,
+                                     "Snow_age",
+                                     d4start, d4count, ivar);
```
jhamman: As mentioned in https://github.com/UW-Hydro/tonic/pull/43#discussion_r57246936, lets change this variable to a double and have the units be in days. There's no reason it should have any dependence on the model timestep. This will require a change in a few places throughout the code.

tbohn: So you want me to make this change as part of this (and [UW-Hydro/tonic#43](https://github.com/UW-Hydro/tonic/pull/43)) PR?

I'd prefer to make that a separate issue, since it's going to affect the classic state file i/o and the snow_utility.c and lakes.eb.c code too.  The purpose of this PR was simply to bring the image driver up to the same functionality as the classic driver in terms of state file i/o.

tbohn: Just to be clear: I'd be happy to make the code changes about last_snow that you're suggesting.  I just don't think it should be part of this PR.

jhamman: okay, I'll open a new issue.

jhamman: see https://github.com/UW-Hydro/VIC/issues/421

tbohn:
OK, I've merged the latest commits from develop into this branch and resolved merge conflicts.  So, it's ready for a more thorough review.

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1009 @@
...
+    dimlen = get_nc_dimension(filenames.init_state, "ni");
+    if (dimlen != global_domain.n_nx) {
+        log_err("Number of grid columns in state file does not "
+                "match parameter file");
+    }
+    dimlen = get_nc_dimension(filenames.init_state, "nj");
+    if (dimlen != global_domain.n_ny) {
+        log_err("Number of grid rows in state file does not "
+                "match parameter file");
+    }
```
jhamman: `"ni"` and `"nj"` should be replaced with `global_domain->info.x_dim` and `global_domain->info.y_dim`. We do not want to require that ni/nj are used as dimensions in the domain file and we do want to check that the state file matches the domain file.

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1009 @@
...
+    status = nc_inq_varid(nc.nc_id, "lon", &lon_var_id);
+    if (status != NC_NOERR) {
+        log_err("Unable to find variable \"lon\" in %s",
+                filenames.init_state);
+    }
+    status = nc_inq_varid(nc.nc_id, "lat", &lat_var_id);
+    if (status != NC_NOERR) {
+        log_err("Unable to find variable \"lat\" in %s",
+                filenames.init_state);
+    }
```
jhamman: Again, these should come from `global_domain->info`

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1009 @@
...
+        dvar = calloc(global_domain.n_nx, sizeof(*dvar));
+        d1count[0] = global_domain.n_nx;
+        status = nc_get_vara_double(nc.nc_id, lon_var_id,
+                                    d1start, d1count, dvar);
+        if (status != NC_NOERR) {
+            log_err("Error reading data from \"lon\" in %s",
+                    filenames.init_state);
+        }
+        for (i = 0; i < global_domain.n_nx; i++) {
+            if (dvar[i] != (double) global_domain.locations[i].longitude) {
```
jhamman: For all these comparisons, I think it would be better to compare these floating points using this function: https://github.com/UW-Hydro/VIC/blob/develop/vic/vic_run/src/comparisons.c#L48

vic/drivers/shared_image/src/vic_restore.c
tbohn: You mean lon/lat_var_id?  Wouldn't these be potentially different from file to file, i.e. wouldn't global_domain->info's versions of them be coming from some other file like the forcings or parameters?  I'm assuming these ids reflect the order in which they're defined in the file at the time of file creation.  If someone creates an initial state file outside of VIC or tonic, are we going to penalize them for not having defined their dimensions in the exact same order as VIC?  Or am I missing something here?

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1009 @@
...
+    // soil thermal node depths
+    dvar = calloc(options.Nnode, sizeof(*dvar));
+    get_nc_field_double(filenames.init_state, "node_depth",
+                        d1start, d1count, dvar);
+    for (i = 0; i < options.Nnode; i++) {
+        if (dvar[i] != soil_con[0].Zsum_node[i]) {
+            log_err("Soil node depths in state file do not match "
+                    "those computed by VIC");
+        }
+    }
```
jhamman: Thanks for putting this function in here, I think its going to save much headache as we go forward.  I think we want to also compare the `mask` variable in the statefile to what we have in the global domain.

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK to the first sentence.  Although we'll have to document somewhere that VIC won't be able to read a state file whose dimension names don't match those of the forcing and parameter files.  Obviously in most cases VIC will create the state file to match, but if someone creates a state file outside of VIC or tonic, they'll need to know that their dimension names must match those of the forcings and parameters that they're using.  Am I understanding this correctly?

What if someone uses forcings from one source (say, NARR) and parameters from some other source (say, Livneh et al 2013).  Will they need to modify the dimension names to all be exactly the same?  (I realize that in mixing/matching like that, probably other modifications would also need to be made to the files anyway).

I wonder if it wouldn't be better to have separate sets of dimension names, ids, etc for forcing, parameter, and state files (and any other input files).  These would all need to be specified by the user in the global parameter file.  Or some logic would have to be inserted into VIC to automatically detect these, but that seems to be too much.

vic/drivers/shared_image/src/vic_restore.c
tbohn: ...but reading, writing and validating the mask is for a different/future PR, right?  (Just making sure)

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1009 @@
...
+            d5start[1] = k;
+            for (j = 0; j < options.Nlayer; j++) {
+                d5start[2] = j;
+                get_scatter_nc_field_double(filenames.init_state,
+                                            "Soil_moisture",
+                                            d5start, d5count, dvar);
+                for (i = 0; i < local_domain.ncells_active; i++) {
+                    v = veg_con_map[i].vidx[m];
+                    if (v >= 0) {
+                        all_vars[i].cell[v][k].layer[j].moist = dvar[i];
```
jhamman: I just want to make sure I understand how the indexing into `dvar` is working. Isn't `dvar` the size of the global domain?  And isn't `all_vars` the size of the `local_domain`?  So how are we handling locations that are not active?

vic/drivers/image/src/vic_populate_model_state.c
```diff
@@ -24,13 +24,15 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *****************************************************************************/
 
+#include <vic_def.h>
+#include <vic_run.h>
 #include <vic_driver_image.h>
```
jhamman: These are included in `vic_driver_image.h`.

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1009 @@
...
+            // slow carbon: tmpval = lake_var.soil.CSlow;
+            get_scatter_nc_field_double(filenames.init_state,
+                                        "Lake_CSlow",
+                                        d2start, d2count, dvar);
+            for (i = 0; i < local_domain.ncells_active; i++) {
+                all_vars[i].lake_var.soil.CSlow = dvar[i];
+            }
+        }
+
+        // snow age: lake_var.snow.last_snow
```
jhamman: Should we drop all this into a `if (options.LAKES) {` block?

vic/drivers/shared_image/src/vic_restore.c
tbohn: Isn't get_scatter_nc_field* doing the mapping between global and local domains?  The for loop over local_domain.ncells_active ensures that the assignment to all_vars doesn't overstep the bounds of all_vars and only accesses the good data from dvar, right?  I've simply taken the same logic used in vic_store() and vic_init() and re-used it.  Unless I made a mistake...

vic/drivers/shared_image/src/vic_restore.c
tbohn: I think you're right - thanks for catching that!

vic/drivers/shared_image/src/vic_restore.c
jhamman: Ah, it didn't register to me that you were using `get_scatter_nc_field_double` and not `get_nc_field_double`. I'm still not entirely sure this is the best way to do this but @bartnijssen will be better at discerning what is best.

vic/drivers/shared_image/src/vic_restore.c
jhamman: No, I mean the `"lat"` and `"lon"` strings.  Since we read and write the statefile ourselves, its not unreasonable to expect that the coordinates, dimensions, and mask match the domain file. It is the domain file that populates `global_domain->info`.

vic/drivers/shared_image/src/vic_restore.c
jhamman: > OK to the first sentence. Although we'll have to document somewhere that VIC won't be able to read a state file whose dimension names don't match those of the forcing and parameter files.

Here we're only talking about the state file.  Currently we don't have any checks in place for the parameter or forcing file conventions. Like I said [here](https://github.com/UW-Hydro/VIC/pull/402/files#r57461119), we are writing the statefile, so we control what the variable and dimension names are.  All I'm saying is that they should match the domain file conventions.

> I wonder if it wouldn't be better to have separate sets of dimension names, ids, etc for forcing, parameter, and state files (and any other input files)

At some point, we may go there, but for now, we're trying to ovoid hard coding the names in VIC and instead relying on the user (actually `vic_store` or `tonic`) to have the state file match the domain file

vic/drivers/shared_image/src/vic_restore.c
jhamman: Sure, opened a new issue: https://github.com/UW-Hydro/VIC/issues/422

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK, will do.

I think i see what you're saying - we're only talking about the state file, and it will have to match the domain file (not the forcing and param files).

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK, will do.

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK, will do.

vic/drivers/image/src/vic_populate_model_state.c
tbohn: OK

jhamman:
Okay, I've reviewed everything here.  @bartnijssen will need to followup with a detailed review as well.

The changes in `vic_store` were so long that github wont show the diff.  I did review that diff on my local machine and I just had one comment.

1.  make sure the writing of the lake variables is in an if block as well.

tbohn:
OK, will do.  Thanks for the review.

vic/drivers/image/src/vic_populate_model_state.c
tbohn: OK, I'll make this change.  I'll also make it to vic_populate_model_state() in cesm driver, since it has the same issue.

vic/drivers/image/include/vic_driver_image.h
```diff
@@ -31,11 +31,12 @@
 
 #define VIC_DRIVER "Image"
 
+void display_current_settings(int);
 void get_forcing_file_info(param_set_struct *param_set, size_t file_num);
 void get_global_param(FILE *);
+void vic_force(void);
```
tbohn: Here I simply ordered the functions alphabetically, so vic_force() changed position.

vic/drivers/image/src/vic_force.c
```diff
@@ -339,7 +339,8 @@ vic_force(void)
             if (vidx != -1) {
                 for (j = 0; j < NF; j++) {
                     if ((veg_hist[i][vidx].fcanopy[j] < MIN_FCANOPY) &&
-                        ((current == 0) || (options.FCAN_SRC == FROM_VEGHIST))) {
+                        ((current == 0) ||
+                         (options.FCAN_SRC == FROM_VEGHIST))) {
```
tbohn: This is just the result of uncrustify.

vic/drivers/shared_image/src/vic_restore.c
tbohn: Oops, it already *is* in an if block, just a long one whose start is several lines above.

tbohn:
Actually, the lake variables in both vic_restore() and vic_store() are all within `if (options.LAKES)` blocks.  So, no change needed for them.

vic/drivers/image/src/vic_populate_model_state.c
bartnijssen: @tbohn - ping me in code review when you have made these changes. I'll hold off on code review till then.

vic/drivers/shared_image/src/vic_restore.c
tbohn: I'm not seeing any examples of assert_close* being used anywhere else in vic, at least not under vic_run or drivers/*...  What would you like to use as values for abs_tol and rtol?

jhamman: agreed.  I don't think we set grid decimal in the image driver (unless its set during initialization). I think you're on the right track with `abs_tol=0.0001` and `rot = 0.`.

tbohn: OK, I'll do that.  Should abs_tol and rtol be local vars or constants in a .h file?  I'm a little reluctant to set up constants since they maybe should have context-dependent values (for example, abs_tol of 0.0001 would be too coarse if we get finer than 1/16 degree resolution).  How about for now, just to get the code functioning, I set up local vars within check_state_file() and then a separate issue and PR can deal with how to make them work in a more general way?

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+    size_t                     d3start[3];
+    size_t                     d4count[4];
+    size_t                     d4start[4];
+    size_t                     d5count[5];
+    size_t                     d5start[5];
+    size_t                     d6count[6];
+    size_t                     d6start[6];
+
+    // validate state file dimensions and coordinate variables
+    check_init_state_file();
```
bartnijssen: Should this return a status? pass/fail - or do you just exit from within that function

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+
+    // read state variables
```
bartnijssen: remove comment since you do not read till later

vic/drivers/shared_image/src/vic_restore.c
bartnijssen: I think if Ted follows the logic in the other places where `get_scatter` is used, than this should work just fine. The whole point is that at this level the mapping should be invisible.

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+                for (i = 0; i < local_domain.ncells_active; i++) {
+                    v = veg_con_map[i].vidx[m];
+                    if (v >= 0) {
+                        all_vars[i].cell[v][k].CLitter = dvar[i];
+                    }
+                }
+            }
+        }
+
+        // intermediate carbon: tmpval = tmpval = cell[veg][band].CInter;
```
bartnijssen: Confusing comment

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+        get_scatter_nc_field_double(filenames.init_state,
+                                    "Lake_ice_temp",
+                                    d2start, d2count, dvar);
+        for (i = 0; i < local_domain.ncells_active; i++) {
+            all_vars[i].lake_var.tempi = dvar[i];
+        }
+
+        // lake ice snow water equivalent: lake_var.swe
+        get_scatter_nc_field_double(filenames.init_state,
+                                    "Lake_ice_snow_water_equivalen",
```
bartnijssen: Are we really saving one character by shaving off the final `t` or is this just a typo. Either way, please add the `t`

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+            log_err("Error reading data from \"%s\" in %s",
+                    global_domain.info.lon_var, filenames.init_state);
+        }
+        for (i = 0; i < global_domain.n_nx; i++) {
+            if (!assert_close_double(dvar[i], global_domain.locations[i].longitude, rtol, abs_tol)) {
+                log_err("Longitudes in initial state file do not "
+                        "match parameter file");
+            }
+        }
+        free(dvar);
```
bartnijssen: Since the `calloc` is within an `if` block, the `free()` should be as well.

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+                global_domain.info.lon_var, filenames.init_state);
+    }
+    status = nc_inq_varid(nc.nc_id, global_domain.info.lat_var, &lat_var_id);
+    if (status != NC_NOERR) {
+        log_err("Unable to find variable \"%s\" in %s",
+                global_domain.info.lat_var, filenames.init_state);
+    }
+    if (global_domain.info.n_coord_dims == 1) {
+        d1start[0] = 0;
+        dvar = calloc(global_domain.n_nx, sizeof(*dvar));
```
bartnijssen: Check that `calloc` is successful or `log_err` otherwise.

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+        }
+        for (i = 0; i < global_domain.n_nx; i++) {
+            if (!assert_close_double(dvar[i], global_domain.locations[i].longitude, rtol, abs_tol)) {
+                log_err("Longitudes in initial state file do not "
+                        "match parameter file");
+            }
+        }
+        free(dvar);
+
+        dvar = calloc(global_domain.n_ny, sizeof(*dvar));
```
bartnijssen: check that `calloc` is successful

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+                log_err("Latitudes in initial state file do not "
+                        "match parameter file");
+            }
+        }
+        free(dvar);
+    }
+    else if (global_domain.info.n_coord_dims == 2) {
+        d2start[0] = 0;
+        d2start[1] = 0;
+        dvar = calloc(global_domain.n_ny * global_domain.n_nx, sizeof(*dvar));
```
bartnijssen: check that `calloc` is successful

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -0,0 +1,1011 @@
...
+    for (i = 0; i < options.Nnode; i++) {
+        if (dvar[i] != soil_con[0].dz_node[i]) {
+            log_err("Soil node intervals in state file do not match "
+                    "those computed by VIC");
+        }
+    }
+    free(dvar);
+
+    // soil thermal node depths
+    dvar = calloc(options.Nnode, sizeof(*dvar));
```
bartnijssen: check `calloc` is successful

vic/drivers/shared_image/src/vic_restore.c
tbohn: Do you have a preference?  Currently we just exit with a log_err() call.

vic/drivers/shared_image/src/vic_restore.c
tbohn: Fixed.

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK

vic/drivers/shared_image/src/vic_restore.c
jhamman: Exiting from within `check_init_state_file()` is my preference.  The only reason to return a status is if we want to do something other than exit.

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK, will do both here and elsewhere.

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK

vic/drivers/shared_image/src/vic_restore.c
tbohn: OK

PR #425, opened 2016-03-25
Miscellaneous lake fixes

### LINKED ISSUE #405: Incorrect aerodynamic resistance over lakes
I've already identified the problem and the fix is two lines in lakes.eb.c.

### LINKED ISSUE #406: Inappropriate choice for default initial lake temperature in absence of initial state file
Currently, in the absence of an initial state file, VIC initializes the entire lake water column to the first time step's air temperature.  A common case when running simulations with lakes (yes, multiple groups are running VIC with lakes) even in temperate regions is that the simulation starts at midnight, Jan 1, and often air temperature is below freezing.  Initialize_model_state() does put a lower bound of 0 C for the initial lake water temperature, but this still is too cold, since it will result in a lake that is frozen solid, which rarely happens in real life.  A more appropriate lower bound would be 4 C, since lake water tends to resist cooling below 4 C due to water's temperature-density relationship at those temperatures.

Changing the lower bound would require a one-line fix.

Note that an even better solution would be to initialize the lake water temperature to annual average air temperature.  This could be possible if annual average air temperature has been supplied in the soil parameter file.  We could check whether it has been supplied, and use it if so, otherwise use temperature on the first time step with a lower bound of 4 C.  Any feedback on this is welcome.

### LINKED ISSUE #419: Incorrect albedo reported over lakes
VIC's internally-used albedo over lakes is fine.  But it isn't being reported correctly in output files.  This won't affect results in any way (except for reported albedos).  The fix is trivial - a one-liner.

### LINKED ISSUE #420: Incorrect energy fluxes reported over lakes
Similar to issue #419 for albedo, several energy flux terms are reported incorrectly over lakes.  They've been scaled by the wrong area fraction.  The fix here is similarly trivial, and, as for albedo, the fix occurs in `lakes.eb.c` under `vic_run`.

tbohn:
This fix addresses several minor bugs.  Several lake processes (aerodynamic resistance, albedo, latent/sensible heat fluxes, net radiation, etc) were reported incorrectly or not at all in output files.  The fixes for those were all minor (1-liners).  Lake temperatures were initialized to unrealistic temperatures in the absence of an initial state file.  The fix for that was a little more involved: replaced the use of first step's air temperature with use of annual average soil temperature.

closes #405 
closes #406 
closes #419 
closes #420

### DIFF
--- a/docs/Development/ReleaseNotes.md
+++ b/docs/Development/ReleaseNotes.md
@@ -121,6 +121,11 @@ This is a major update from VIC 4. The VIC 5.0.0 release aims to have nearly ide
 
 #### Bug Fixes:
 
+1. Miscellaneous fixes to lake module
+
+	Several lake processes (aerodynamic resistance, albedo, latent/sensible heat fluxes, net radiation, etc) were reported incorrectly or not at all in output files. This has been fixed. In addition, in the absence of an initial state file, lake temperatures were initialized to unrealistic temperatures (the air temperature of the first simulation time step). To fix this, we now initialize the lake temperature to annual average soil temperature.
+
+
 ------------------------------
 
 ## VIC 4.2.c [![DOI](https://zenodo.org/badge/doi/10.5281/zenodo.35302.svg)](http://dx.doi.org/10.5281/zenodo.35302)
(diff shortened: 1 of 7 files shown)


jhamman:
No problems with the code changes.  We need to add a comment in the [Release Notes](https://github.com/UW-Hydro/VIC/blob/develop/docs/Development/ReleaseNotes.md#bug-fixes) describing the bugs fixed here.

tbohn:
Should that be included in this PR?

jhamman:
Yes, doc and code changes go together.

bartnijssen:
@tbohn : you're out of date on this one - please merge `develop` into this branch one more time, resolve conflicts and then I'll go ahead and merge the PRs

PR #439, opened 2016-03-30
Bugfix/vic force wind

wietsefranssen:
Fixed hardcoded netcdf variable name 'wind'
in `vic_force.c` the netcdf variable for wind was still hardcoded. This is now fixed.


jhamman:
This looks good.  Thanks for fixing this.  Can you add a line to this global parameter file so the tests on travis pass: https://github.com/UW-Hydro/VIC/blob/develop/tests/examples/global_param.image.STEHE.feb.txt#L33

```
FORCE_TYPE    WIND           wind
```

wietsefranssen:
@jhamman done

Issue #443, opened 2016-03-31
add udunits support


wietsefranssen:
In order to avoid conversion errors (eg to double conversion) while reading forcing data, but also to make it easier for the user to prepair forcing data with different units, I suggest to add support for the udunits2 library.
This library can later also be used to choose the preferred output units of the variables.

I suggest to start to add lines to read the `units` attribute of the forcing variables from the netcdf file.
Then make a conversion function like `float convert(char from, char to, float data)`
and use this function directly after reading the data from the file to convert the the standard units vic uses for its calculations.

Later we can extend this with more functionality, error handling, vic output etc.

@bartnijssen @jhamman what do you thing about this idea?

wietsefranssen:
I see now that my previous message appeared just a few seconds ago. In the meanwhile I made a working example. I will make a pull request in order to give you insight of what I did. (in case you want an other way of comparing/sharing in stead of making a pull request, just let me know.

PR #444, opened 2016-03-31
Added udunits library and functionality

wietsefranssen:
see #443


jhamman:
To get the CESM and Image driver tests to pass on Travis, you'll need to make sure `udunits` is installed.  Do so by adding `libudunits2-dev` to the list of [`apt_packages`](https://github.com/UW-Hydro/VIC/blob/develop/.travis.yml#L14) in the `.travis.yml` file.

vic/drivers/shared_image/src/vic_udunits.c
```diff
@@ -0,0 +1,59 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *****************************************************************************/
+
+#include <vic_driver_shared_image.h>
+#include <udunits2.h>
```
jhamman: include in `vic_driver_shared_image.h` so you just have to include it once.

vic/drivers/image/src/vic_force.c
```diff
@@ -25,6 +25,7 @@
  *****************************************************************************/
 
 #include <vic_driver_image.h>
+#include <udunits2.h>
```
jhamman: include in `vic_driver_shared_image.h` so you don't need to include it here.

vic/drivers/shared_image/include/vic_driver_shared_image.h
```diff
@@ -204,5 +205,6 @@ void vic_restore(void);
 void vic_start(void);
 void vic_store(void);
 void vic_write(void);
+cv_converter*udunits_conversion(const char*const from, const char*const to);
```
jhamman: I think uncrustify will reformat this line.  Did you run it yet?

vic/drivers/shared_image/src/vic_udunits.c
wietsefranssen: okay I will do that

vic/drivers/shared_image/include/vic_driver_shared_image.h
wietsefranssen: Yes I did and just tried it again (without succes). No I changed it by hand to: 
`cv_converter *udunits_conversion(const char *const from, const char *const to);`

jhamman:
Thanks @wietsefranssen - 

A few things for @bartnijssen and I to discuss further.

1. Is it okay to have udunits as a dependency for the image driver?
2. Does the CESM driver need/support udunits? Grepping through the source code in the RASM repository does not show any usage of udunits so I think the answer is no. So we'll need to move this code from `drivers/shared_image` to `drivers/image` but the implementation could remain the same.

jhamman:
CESM does not support `udunits` so we'll need to move this to the individual driver level (e.g. `drivers/image/`.

.travis.yml
```diff
@@ -22,6 +23,7 @@ matrix:
     env: TESTID='cesm'
     addons:
       apt_packages:
+        - libudunits2-dev
```
jhamman: remove from cesm driver build

vic/drivers/shared_image/src/vic_udunits.c
```diff
@@ -0,0 +1,58 @@
+/******************************************************************************
```
jhamman: move this file to `drivers/image`

vic/drivers/shared_image/include/vic_driver_shared_image.h
```diff
@@ -31,6 +31,7 @@
 #include <vic_mpi.h>
 
 #include <netcdf.h>
+#include <udunits2.h>
```
jhamman: move to `drivers/image`

vic/drivers/shared_image/include/vic_driver_shared_image.h
```diff
@@ -204,5 +205,6 @@ void vic_restore(void);
 void vic_start(void);
 void vic_store(void);
 void vic_write(void);
+cv_converter *udunits_conversion(const char *const from, const char *const to);
```
jhamman: move to `drivers/image`

jhamman:
@wietsefranssen - this looks great.  The final piece here is just to complete the application for the rest of the forcing variables.

wietsefranssen:
@jhamman thanks Joe. I will complete it for the other forcing variables and commit it in 5minutes.

I think that it is not the final piece yet: we need a solution for the conversion between [mm] and [kg m-2] (for eg precip). I also need to add some extra error handling.

One suggestion/question: What about checking and defining the conversion for the forcing data once at the start of VIC, before going into the nrecs loop?

vic/drivers/image/src/vic_force.c
```diff
@@ -59,6 +59,9 @@ vic_force(void)
     size_t                     d4count[4];
     size_t                     d4start[4];
 
+    cv_converter              *converter;
```
jhamman: is this basically a double?  if so, we may be able to put this in `param_set_struct` so we only have to check the units and generate the converter once.

vic/drivers/image/src/vic_udunits.c
```diff
@@ -0,0 +1,58 @@
...
+ * @brief    udunits2 support.
+ *****************************************************************************/
+cv_converter*
+udunits_conversion(const char*const from,
+                   const char*const to)
+{
+    ut_system   *unitSystem;
+    ut_unit     *ut_from;
+    ut_unit     *ut_to;
+    cv_converter*converter;
```
jhamman: missing a space here.

jhamman:
> we need a solution for the conversion between [mm] and kg m-2

what's the issue here? and for what variable?

vic/drivers/image/src/vic_force.c
wietsefranssen: cv_converter is not a double (a union I assume), but we can still add it to param_set_struct like: `    cv_converter converter[N_FORCING_TYPES];`, check the units and generate the converter once. This is my preference in terms of speed (but this is probably a tiny improvement), but also to avoid repetitious warnings, etc.

wietsefranssen:
@jhamman eg for precipitation. Sometimes it is given in [mm s-1] (or [mm day-1]) and sometimes in [kg m-2 s-1]. if vic uses [kg m-2 s-1] and the netcdf units is given as [mm s-1] (or [mm day-1]) udunits handle this because this [kg m-2]-->[mm] is not a direct conversion (you need to take the density of water into account). 
A simple solution would be to force the user to always use [kg m-2] as units in the netcdf file. 
An other solution would be to handle this kind of exception in de code (in vic_udunits.c). The the user is free choose the unit and VIC takes care of it.

wietsefranssen:
Hi, I have the feeling that this pull request can be closed. We use standard units in VIC and we want as less as possible extra in VIC 5.

Issue #398, opened 2016-03-11
Likely redundant information in state files


tbohn:
At some point it would probably be good to attempt to reduce the number of variables stored in the state files.  There are a large number currently, particularly lake state variables, and I think several of them could contain redundant information or could be derived from the others (i.e. not independent of each other).  Eliminating some of these would save space and make code maintenance easier.

Some good candidates to investigate include:

1. Lake snow-related variables - the lake has its own snow data structure to store information about the snowpack on top of lake ice.  But the lake_var structure also contains some snow-related information.  Some examples are: surf_temp, surf_water, pack_temp, and pack_water, coldcontent, SAlbedo, sdepth.  Since the lake's snow_data structure contains a complete description of the snowpack state, it seems that the lake_var versions of these variables could be eliminated and just re-computed after reading the state file.
2. Lake vertical average temperature - could be derived from individual lake layer temperatures.
3. Lake surface, sarea, and volume - could be derived from lake depth and lake parameters.
4. Soil temperature profile information: In the classic driver, dz_node and node_depth arrays are stored once per grid cell, despite their being constant across the entire domain.
5. Nveg, nbands - in the classic driver, these are stored but are redundant with their counterparts in the veg_param and snow_bands files.  They were (presumably) included so that the ascii state file could be parsed by a script without having to read in the veg_param and snow_bands files.  But not only could space be saved by omitting them, their presence opens up the possibility of discrepancies with the veg_param and snow_bands versions of these.  Although one could argue that keeping them would allow VIC to throw up an error if they differ from the veg_param and snow_bands versions.

tbohn:
Note that I'm not suggesting that this is a high priority for VIC 5.0.  I'm just documenting it.

bartnijssen:
We should be careful to sort this out at some point. We definitely do not want redundant inf. Not just for file size, but because we should not carry copies of the state internally either. This will make it too easy for bugs to creep in. For example, someone would update one and forget that there is another instance.

tbohn:
Regarding the redundant lake state variables: it may be a good idea to remove some or all of those from VIC altogether, not just from the state file.  Particularly the variables pertaining to the snowpack on top of lake ice - there are versions in the lake_var.snow structure, which I would recommend keeping, and in the lake_var structure itself, which I would recommend removing (replacing throughout the lake code with the lake_var.snow versions).

PR #464, opened 2016-04-08
Fix to state file i/o problems and refactoring of initialization

### LINKED ISSUE #424: Inappropriate choice for default initial soil temperature in absence of initial state file
Currently, the default initial soil temperature in the absence of an initial state file is set to the air temperature of the first time step.

A better choice (closer to equilibrium) would be soil_con.avg_temp, as long as this has been supplied.  This would reduce the amount of time necessary to spin up to a usable initial state.  This would also provide a potential solution to the TBD in vic_populate_model_state().

### LINKED ISSUE #457: image driver miscellaneous issues with state file i/o
My previous PR (#402) didn't get fully tested due to problems running the model in image mode.  Now that the model is working, I've found a few problems with state file i/o:

1. call to vic_store() was commented out due to not working with MPI
2. state file hard-coded to be written at end of simulation
3. state file name has date appended twice
4. errors attempting to write lat/lon values
5. errors attempting to write all other coordinate variables
6. lat/lon values read by vic_restore() didn't match those from parameter files

The fixes for these issues are almost complete and include:
To address issues 2 and 3:

- implementing a new function, image/src/check_save_state_flag(), which compares current date to state date and returns true if they're the same, false otherwise (a different version of this function can be implemented in the cesm driver if desired); this will replace the condition `if (current == global_param.nrecs - 1)`
- adding global_param.statesec in vic_def.h and including the statesec value in the output state file name (as is already being done in vic_store() using dmy[current].daysecond)
- modifying get_global_param.c in both classic and image drivers to look for a STATESEC value in addition to STATEYEAR, STATEMONTH, and STATEDAY
- modifying the sample global parameter files in classic and image drivers to add a  STATESEC entry
- adding STATESEC to classic and image display_current_settings.c

To address issues 4 and 6:

- moving logic of initialize_state_file() to vic_store() and eliminating initialize_state_file() - this function is actually unnecessary since the initialization and writing all occur at the same time, and as it currently is, it's splitting file definition-related commands between the two functions, thus preventing lat/lon values from being written
- changing the lat/lon indexing for the 1-dimensional case in both vic_store() and vic_restore(), since a stride is needed in looping over the global_domain.location[] array

To address issue 5:

- for 1-d coordinate variables like "veg_class" and "dz_node", replacing calls to gather_put_nc_field_xxx(), which assumes the data are spatial fields that must be masked, with calls to put_nc_field_xxx()

To address issue 1:

- uncommenting call to vic_store() in vic_image()
- enclosing all dimension and coordinate variable definition and writing commands in an `if (mpi_rank == 0)` block

tbohn:
Fix to state file i/o problems and refactoring of initialization into separate functions for generating a default state and for handling derived state variables.  Most of these functions are shared across drivers.

closes #457 
closes #424 

Quick summary of the changes:

1. Fixes for miscellaneous state file i/o issues: e.g., lat/lon indexing was wrong for the 1d case in both vic_store() and vic_restore(); some other variables were being written incorrectly.
2. Refactoring of state initialization: I made the classic driver look more like the image and cesm drivers, with functions devoted to generating the default state, and handling derived state variables, shared across all drivers.

New structure:

read_soilparam(classic) or vic_init(image/cesm):

- reads soil parameters, does computation of derived soil parameters and properties
- moved the setting of soil thermal node spacing to here
- (image only) calls initialize_* functions

vic_populate_model_state(all drivers):

- (classic only) calls initialize_* functions
- if (options.INIT_STATE): reads initial state file, via read_initial_model_state(classic) or vic_restore(image/cesm); moved the capping of initial soil moisture to max_moist to inside read_initial_model_state() (classic only) - I can add similar code to vic_restore() if desired
- else:generate_default_state(all) and generate_default_lake_state(all) // initializes soil/lake water storage and soil/lake temperatures to default values
- compute_derived_state_vars(all) and compute_derived_lake_dimensions(all) // sets soil thermal node properties and ice contents, and computes lake dimensions that depend on lake storage


tbohn:
Note: I've tested that both classic and image drivers now successfully save state at the appointed time, and either read the specified state file or start from a default state as requested.

Note: results are a little different between the new classic and 4.2.  Not hugely, but still, not the same.  Hopefully this can be investigated further in restart tests.

tbohn:
Another note: strangely, starting from the the default state does not generate root_brent errors anymore after this PR is merged.  However, starting from a state file still does.  Not sure why.

vic/drivers/classic/src/read_soilparam.c
```diff
@@ -692,6 +696,108 @@ read_soilparam(FILE *soilparam,
             temp.Ws = temp.Ws / temp.max_moist[layer];
         }
 
+        // Soil thermal node thicknesses and positions
```
tbohn: This section was originally in the old initialize_model_state() function, but it really should go here, as node spacing is completely determined from soil parameters and global param options.  This way, all soil_con values are populated by the time we exit read_soilparam().

vic/drivers/shared_image/src/vic_init.c
```diff
@@ -792,6 +796,120 @@ vic_init(void)
 
         soil_moisture_from_water_table(&(soil_con[i]), options.Nlayer);
 
+        // Soil thermal node thicknesses and positions
```
tbohn: This section was originally in the old initialize_model_state() function, but it really should go here, as node spacing is completely determined from soil parameters and global param options.  This way, all soil_con values are populated by the time we exit vic_init().

vic/drivers/classic/src/read_initial_model_state.c
```diff
@@ -688,4 +692,90 @@ read_initial_model_state(FILE            *init_state,
             }
         }
     }
+
+    /******Check that soil moisture does not exceed maximum allowed************/
```
tbohn: I simply moved this section from the old initialize_model_state() function.  This way, the new vic_populate_model_state() function on the classic side is much more analogous to the same function on the shared_image side.

vic/drivers/classic/src/get_global_param.c
```diff
@@ -1397,32 +1400,36 @@ get_global_param(FILE *gp)
         if (global_param.stateyear == 0 || global_param.statemonth == 0 ||
             global_param.stateday == 0) {
             log_err("Incomplete specification of the date to save state "
-                    "for state file (%s).  Specified date (yyyy-mm-dd): "
-                    "%04d-%02d-%02d Make sure STATEYEAR, STATEMONTH, and "
-                    "STATEDAY are set correctly in your global parameter "
+                    "for state file (%s).\nSpecified date (yyyy-mm-dd-hh): "
+                    "%04d-%02d-%02d-%05u\nMake sure STATEYEAR, STATEMONTH, "
+                    "and STATEDAY are set correctly in your global parameter "
```
jhamman: You have `-hh` but this is seconds here.  Maybe `-sssss` is better?

vic/drivers/classic/src/get_global_param.c
```diff
@@ -1397,32 +1400,36 @@ get_global_param(FILE *gp)
...
+            global_param.statemonth > MONTHS_PER_YEAR ||
+            global_param.stateday < 1 || global_param.stateday > 31 ||
+            global_param.statesec > SEC_PER_DAY) {
+            log_err("Unusual specification of the date to save state "
+                    "for state file (%s).\nSpecified date (yyyy-mm-dd-hh): "
+                    "%04d-%02d-%02d-%05u\nMake sure STATEYEAR, STATEMONTH, "
+                    "STATEDAY and STATESEC are set correctly in your global "
+                    "parameter file.", filenames.statefile,
+                    global_param.stateyear, global_param.statemonth,
+                    global_param.stateday, global_param.statesec);
```
jhamman: Same comment as above on seconds vs. hours.

vic/drivers/classic/src/read_initial_model_state.c
jhamman: Is it possible to put this in `drivers/shared` then?

tbohn: Put which function in drivers/shared read_initial_model_state or vic_populate_model_state?  And do you mean shared_all or shared_image?

vic_populate_model_state() has some differences across drivers - cesm is different from image, let alone classic.  And read_initial_model_state() is not really shareable...

Or do you mean this section of code that I moved?

jhamman: Well you said that the `vic_populate_model_state()` is analogous to the `image` driver side.  I'm wondering if we want to pull out the common pieces? I'm trying to do a side by side comparison of the two implementations to assess that possibility.

vic/drivers/classic/src/read_initial_model_state.c
```diff
@@ -688,4 +692,90 @@ read_initial_model_state(FILE            *init_state,
...
+                        cell[veg][band].layer[lidx].moist) {
+                        cell[veg][band].layer[lidx].ice[frost_area] =
+                            cell[veg][band].layer[lidx].moist;
+                    }
+                }
+            }
+        }
+
+        // Override possible bad values of soil moisture under lake coming from state file
+        // (ideally we wouldn't store these in the state file in the first place)
```
jhamman: Do we have an open issue for this case (ideally we wouldn't ...).

vic/drivers/classic/src/read_initial_model_state.c
```diff
@@ -688,4 +692,90 @@ read_initial_model_state(FILE            *init_state,
...
+                        lake_var->soil.layer[lidx].moist) {
+                        lake_var->soil.layer[lidx].ice[frost_area] =
+                            lake_var->soil.layer[lidx].moist;
+                    }
+                }
+            }
+        }
+    }
+
+    /******Check that snow pack terms are self-consistent************/
```
jhamman: general comment - can we update the comment format to either `// Check that ...` or `/* Check that ... */`. I think these extra `*`s just clutter things.  And, yes, I know these were already in the code but know is a good time to update them...

vic/drivers/classic/src/vic_populate_model_state.c
```diff
@@ -0,0 +1,101 @@
...
+ * this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *****************************************************************************/
+
+#include <vic_driver_classic.h>
+
+/******************************************************************************
+ * @brief    Initialize the model state (energy balance, water balance, and
+ * snow components).
+ *****************************************************************************/
```
jhamman: indent to align with "Initialize"

vic/drivers/image/src/check_save_state_flag.c
```diff
@@ -0,0 +1,49 @@
...
+ * this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *****************************************************************************/
+
+#include <vic_driver_image.h>
+
+/******************************************************************************
+ * @brief   Function to check whether model state should be saved for the
+ * current time step
+ *****************************************************************************/
```
jhamman: indent to align with "Function"

vic/drivers/image/src/get_global_param.c
```diff
@@ -1168,35 +1171,39 @@ get_global_param(FILE *gp)
...
-                    "STATEDAY are set correctly in your global parameter "
-                    "file.", filenames.statefile, global_param.stateyear,
-                    global_param.statemonth, global_param.stateday);
+            global_param.statemonth > MONTHS_PER_YEAR ||
+            global_param.stateday < 1 || global_param.stateday > 31 ||
+            global_param.statesec > SEC_PER_DAY) {
+            log_err("Unusual specification of the date to save state "
+                    "for state file (%s).\nSpecified date (yyyy-mm-dd-hh): "
+                    "%04d-%02d-%02d-%05u\nMake sure STATEYEAR, STATEMONTH, "
+                    "STATEDAY and STATESEC are set correctly in your global "
```
jhamman: same comment as above on the hours vs. seconds in the comment.

vic/drivers/classic/src/read_initial_model_state.c
tbohn: Problem is that the functions that vic_populate_model_state() calls *are* shared across drivers, but to do so requires looping over the grid cells within the non_classic versions of vic_populate_model_state().

vic/drivers/classic/src/read_initial_model_state.c
tbohn: Yes, #398

vic/drivers/classic/src/get_global_param.c
tbohn: OK.

vic/drivers/classic/src/get_global_param.c
tbohn: OK

vic/drivers/classic/src/read_initial_model_state.c
tbohn: OK

vic/drivers/classic/src/vic_populate_model_state.c
tbohn: OK

vic/drivers/image/src/check_save_state_flag.c
tbohn: OK

vic/drivers/image/src/get_global_param.c
tbohn: OK

vic/drivers/shared_all/src/compute_derived_state_vars.c
```diff
@@ -0,0 +1,234 @@
...
+                                             soil_con->Zsum_node,
+                                             energy[veg][band].T,
+                                             options.Nnode);
+                    }
+                }
+            }
+        }
+    }
+
+    return(0);
```
jhamman: we've been moving away from returning an exit status unless it will be used later.  In this case, we should exit immediately if there are any errors and we can return with `void` instead.

vic/drivers/shared_all/src/generate_default_lake_state.c
```diff
@@ -0,0 +1,61 @@
...
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *****************************************************************************/
+
+#include <vic_driver_shared_all.h>
+
+/******************************************************************************
+ * @brief    Initialize the lake model state (energy balance, water balance,
+ * and snow components) to default values.
```
jhamman: indent second line of comment.

vic/drivers/shared_all/src/generate_default_state.c
```diff
@@ -0,0 +1,113 @@
...
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *****************************************************************************/
+
+#include <vic_driver_shared_all.h>
+
+/******************************************************************************
+ * @brief    Initialize the model state (energy balance, water balance, and
+ * snow components) to default values.
```
jhamman: indent.

vic/drivers/shared_all/src/initialize_lake.c
```diff
@@ -149,14 +63,34 @@ initialize_lake(lake_var_struct  *lake,
...
+    lake->pack_temp = 0.0;
+    lake->pack_water = 0.0;
+    lake->SAlbedo = 0.0;
+    lake->sdepth = 0.0;
+    lake->surf_temp = 0.0;
+    lake->surf_water = 0.0;
+    lake->swe = 0.0;
+    for (i = 0; i < MAX_LAKE_NODES; i++) {
+        lake->temp[i] = 0;
+    }
```
jhamman: consider also setting `lake->temp` in the for loop at line 75 when setting `density`.

vic/drivers/shared_image/src/vic_init.c
jhamman: Just out of curiosity. What happens if we write an state file with one set of thermal node properties/dimensions and try to use that initial state to startup a new simulation with a different set of thermal node properties/dimensions?  Do we raise an error if the run settings do not match the statefile?

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -945,9 +947,13 @@ check_init_state_file(void)
...
-        for (i = 0; i < global_domain.n_ny; i++) {
-            if (!assert_close_double(dvar[i],
-                                     global_domain.locations[i].latitude, rtol,
+        // implicitly nested loop over ni and nj with i set to 0;
+        // j stride = n_nx
+        for (j = 0; j < global_domain.n_ny; j++) {
+            if (!assert_close_double(dvar[j],
+                                     global_domain.locations[j *
+                                                             global_domain.n_nx]
+                                     .latitude, rtol,
```
jhamman: So this just checks the latitudes in the furthest left (west) grid cells?  Why not loop over `global_domain.ncells` and use the `io_index` to check all coordinate locations?

vic/vic_run/src/compute_derived_lake_dimensions.c
```diff
@@ -0,0 +1,108 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *****************************************************************************/
+
+#include <vic_driver_shared_all.h>
+#include <vic_run.h>
```
jhamman: `<vic_run.h>` is included in `vic_driver_shared_all.h`

vic/vic_run/src/compute_derived_lake_dimensions.c
```diff
@@ -0,0 +1,108 @@
...
+                "volume = %e", 0, depth, tmp_volume);
+        return(status);
+    }
+    else if (status > 0) {
+        log_err("Warning in get_volume: lake depth exceeds maximum; "
+                "setting to maximum; record = %d", 0);
+    }
+    lake->volume = tmp_volume + lake->ice_water_eq;
+
+    return(0);
```
jhamman: return `void`

vic/vic_run/src/lakes.eb.c
```diff
@@ -25,6 +25,7 @@
  *****************************************************************************/
 
 #include <vic_run.h>
+#include <vic_driver_shared_all.h>
```
jhamman: `<vic_run.h>` is included in `vic_driver_shared_all.h`

jhamman:
Nice work @tbohn.  Looks good.  @yixinmao - you're up!

vic/drivers/shared_image/src/vic_restore.c
tbohn: This is for the 1-d lat/lon case - as determined by the grid domain file, not the state file.  Accordingly, if it's the 1-d case, then all lats will match the furthest west ones (in the grid from the grid domain file, parameter file, and forcings, since they are all consistent, right?).  Meanwhile, again because its the 1-d case, there is only 1 dimension to check in the state file's lat var - one lat value per grid row.  We already checked earlier whether the dimensionality of the lat and lon variables was what it should be.  So this check should be sufficient.  At least I think so.

vic/drivers/shared_image/src/vic_init.c
tbohn: There's already a check for that in vic_restore.c, line 1012 - a mismatch raises an error.

vic/drivers/shared_all/src/initialize_lake.c
tbohn: I was just keeping the variables in alphabetical order.

vic/drivers/shared_all/src/compute_derived_state_vars.c
tbohn: OK

vic/drivers/shared_all/src/generate_default_lake_state.c
tbohn: OK

vic/drivers/shared_all/src/generate_default_state.c
tbohn: OK

vic/vic_run/src/compute_derived_lake_dimensions.c
tbohn: OK

vic/drivers/shared_image/src/vic_restore.c
jhamman: sounds good.

vic/drivers/classic/src/vic_populate_model_state.c
```diff
@@ -0,0 +1,101 @@
+/******************************************************************************
+ * @section DESCRIPTION
+ *
+ * This routine initializes the model state (energy balance, water balance, and
+ * snow components).
+ *
+ * If a state file is provided to the model than its
```
yixinmao: "then" instead of "than"

vic/drivers/classic/src/read_initial_model_state.c
```diff
@@ -42,29 +42,33 @@ read_initial_model_state(FILE            *init_state,
                          int              Nveg,
                          int              Nbands,
                          int              cellnum,
-                         soil_con_struct *soil_con)
+                         soil_con_struct *soil_con,
+                         lake_con_struct  lake_con)
 {
-    extern option_struct options;
+    extern option_struct     options;
```
yixinmao: Smaller space?

vic/drivers/classic/src/read_initial_model_state.c
```diff
@@ -688,4 +692,90 @@ read_initial_model_state(FILE            *init_state,
...
+                        lake_var->soil.layer[lidx].moist) {
+                        lake_var->soil.layer[lidx].ice[frost_area] =
+                            lake_var->soil.layer[lidx].moist;
+                    }
+                }
+            }
+        }
+    }
+
+    // Check that snow pack terms are self-consistent
```
yixinmao: Ideally we wouldn't write bad snow states also, right?

vic/drivers/classic/src/vic_populate_model_state.c
```diff
@@ -0,0 +1,101 @@
...
+    if (options.LAKES) {
+        tmp_lake_idx = lake_con.lake_idx;
+        if (tmp_lake_idx < 0) {
+            tmp_lake_idx = 0;
+        }
+        initialize_lake(lake, lake_con, soil_con, &(cell[tmp_lake_idx][0]),
+                        false);
+    }
+    initialize_energy(energy, Nveg);
+
```
yixinmao: Currently, in the following, `if (options.INIT_STATE)` reads in initial state file and modifies some direct states (i.e., states that cannot be derived and must be saved), supposedly; `else` if no initial state, then a part of states are modified. Then `compute_derived_state_vars` again modifies a part of states that are derived. Then the rest of the states might be derived and modified in the vic_run core code.

To me this seems a bit confusing because it's not very clear which state variables are 'direct' and must be saved and read from initial state file, and which are 'derived'; and it is possible that some 'direct' states are not correctly saved and read, causing inexact restart (as I'm seeing now for a simple bare soil case). Would it be possible to separate 'direct' and 'derived' states in a cleaner way? For example, maybe not call `initialize_*` functions at all, and then in `if (options.INIT_STATE)`, read in all 'direct' states from state file (as it is doing now supposedly); in `else`, set the same set of 'direct' states to their default values (I guess currently many of the default values are zero, so they are assigned above in `initialize_*`). In this way, all the 'derived' variables will either be assigned in `compute_derived_state_vars`, or in the core `vic_run` computation; but if any 'derived' variables are used before being assigned, then there would be an error and we should avoid it from happening.

These are just some thoughts based on my understanding of the code structure, and not sure if makes sense at all...

tbohn: @yixinmao - I can't say what the cause of the inexact restarts may be.  But to my knowledge, the derived variables are not saved in the state file.  All state variables are initialized to 0 in the `initialize_*` functions (e.g., `initialize_soil()`).  Then, some are overwritten by either reading the state file or generating the default state, and others are subsequently overwritten by `create_derived*` (this is the same set of variables regardless of whether a state file was read or a default state was set).  So if there are any uninitialized variables, they simply need to be added to the `initialize*` functions.

Yes, the set of variables that are assigned values in `generate_default*` is smaller than the set of variables stored in the state files.  This is because only those variables that need to have non-zero values are set in `generate_default*`.

There is one main exception to this: there are many lake state variables saved in the state file that are actually redundant, i.e. can be derived from the others.  There already is an open issue about this but I think it is slated for 5.1 as it is not essential to getting 5.0 running correctly.

vic/drivers/classic/src/read_initial_model_state.c
bartnijssen: `uncrustify` should take care of this. Make sure the code is `uncrustified` before it gets merged.

vic/drivers/classic/src/vic_populate_model_state.c
bartnijssen: Isn't that basically the difference between `prognostic` and `diagnostic` variables? Ideally we should make sure to only store `prognostic` variables and recalculate the `diagnostic` variables, so that they are consistent with the `prognostic` ones. Please correct me if I am wrong.

bartnijssen:
@tbohn : merge conflicts (probably because of Joe's PR that was just merged) - please merge `develop` into this PR

vic/drivers/classic/src/read_initial_model_state.c
tbohn: But I did run uncrustify...  that change happened because of uncrustify...

vic/drivers/classic/src/read_initial_model_state.c
tbohn: True.  I copied this code from its original position outside of read_initial_model_state().  I think this code could be removed without any harm...  let me look into it.

vic/drivers/classic/src/vic_populate_model_state.c
tbohn: I think you're correct.  I think the code is doing essentially what you're describing (with the exception of the lake state variables, which there's an open issue for), although we're not using the same words.

tbohn: So, do we want to have this PR wait until @yixinmao 's restart tests are done (and any necessary bug fixes indicated by them are incorporated into this) or can this PR be merged now and fixes that arise from restart tests can be handled as separate issues?  Those fixes might very well lead to better organization of the variables and code into prognostic and diagnostic variables...

jhamman: I do not think we want to wait on getting exact restarts to merge this. That process could take some time and we don't want to hold up other work.

vic/drivers/classic/src/read_initial_model_state.c
jhamman: This is aligned with the section of variables below.  No need to make further edits.

vic/vic_run/src/lakes.eb.c
jhamman: I just noticed this in my follow up review.  We do not want `vic_driver_shared.h` anywhere in `vic_run`.  If we need to move a function from `drivers/shared_all` to `vic_run`, we can do that but this include needs to go.

vic/drivers/classic/src/read_initial_model_state.c
tbohn: I'm inclined to remove this section of the code.  The cap on soil moisture storage seems more valuable to me, since during calibration the soil moisture storage capacity is commonly modified, which could lead to errors if we didn't do that cap.  But to cap snowpack liquid moisture storage seems to be too much - snowpack liquid water storage capacity is not a commonly-calibrated parameter - if we accept this adjustment, then why not cap all other state variables too?  So, I'll remove it.

vic/drivers/classic/src/vic_populate_model_state.c
tbohn: OK

vic/drivers/classic/src/read_soilparam.c
yixinmao: It seems that currently the last section (around the last ~15 lines) in `initialize_soil.c` also derives some `soil_con` values. Should that part be moved here to `read_soilparam()` also?

tbohn: Yes, I think you are right.  Good catch.  I need to move those into both read_soilparam() and vic_init().

PR #467, opened 2016-04-10
Fix for soil layer temperatures when soil thermal nodes do not reach bottom of soil column.

### LINKED ISSUE #377: ERROR: Soil thermal nodes do not extend below bottom soil layer
I ran VIC v. 4.2.c over a basin in the Columbia River Basin (PNW, U.S.) in energy balance mode (3-hour time step). The model ran to completion for all cells; however, at least one cell gave the following error:

"ERROR: Soil thermal nodes do not extend below bottom soil layer, currently unable to handle this condition."

I have 10 soil thermal nodes and 3 soil layers. FROZEN_SOIL is FALSE. Damping depth = 4 and total soil depth = 4.0378. This error message is confusing and probably unnecessary when FROZEN_SOIL=FALSE.

tbohn:
This change is relatively minor.  This addresses #377, at least for 5.0.  The error in #377 was reported for 4.2.c, even when FROZEN_SOIL was FALSE.  I was not able to reproduce this error in 5.0 unless FROZEN_SOIL was TRUE.  What happens is: if the bottom of the lowest soil layer is deeper than the soil thermal damping depth (dp), and if FROZEN_SOIL is true, then when the model attempts to estimate layer ice content via estimate_layer_ice_content(), it encounters a layer for which soil thermal nodes do not cover the entire layer.  In this case, VIC would abort.

The thing is, if the log_err() call were removed, the logic in estimate_layer_ice_content() would actually work fine: the temperature of the deepest node would be used for all depths below it.  So, my fix here was simply to remove the log_err() call.

closes #377


vic/vic_run/src/soil_conduction.c
```diff
@@ -367,8 +367,10 @@ estimate_layer_ice_content(layer_data_struct *layer,
             max_nidx++;
         }
         if (max_nidx >= Nnodes) {
-            log_err("Soil thermal nodes do not extend below bottom "
-                    "soil layer, currently unable to handle this condition.");
+            // If we get here, soil thermal nodes don't extend all the way
+            // down to the bottom of the lowest layer.  In this case, just
+            // use the deepest node to represent all deeper temperatures.
+            max_nidx = Nnodes - 1;
```
jhamman: Do you think it makes since to raise a warning here instead of an error?

tbohn: Sure, easy enough to do and you're right, the user should be told about it in case it was unintentional.

PR #468, opened 2016-04-11
Removed all instances of month from vic_run().

### LINKED ISSUE #358: Remove month information out of vic_run()
There are still a few instances in which the current month is used (mainly as an array index) in vic_run() to determine parameter values.  Ideally we can move all references to the current month (or date or record) outside of vic_run().

tbohn:
All instances of month have been removed from vic_run().  This has been accomplished in 2 primary ways:

1. displacement and roughness have been added to the veg_con, veg_hist, and veg_var structures, and are now populated in read_vegparam()/vic_init() and vic_force() via simple copying from veg_lib().  The veg_var versions of these terms are now what are accessed in vic_run() rather than the veg_lib versions.
2. coszen has been added to the atmos structure, and is populated via calls to compute_coszen() in vic_force().  coszen is only allocated and populated if CARBON is TRUE.  all other calls to compute_coszen() in the code below vic_run() have been removed.

This technically closes #358 inasmuch as it removes all instances of month from vic_run().  However, issue #358 also mentions other instances of the dmy struct being used within vic_run(), which this PR does not address due to the invasive nature of those occurrences.  So, perhaps issue #358 can be closed but another issue needs to be opened covering these other instances of dmy in vic_run().


jhamman:
The changes here look good.  Ping me when we are passing the CI tests.

vic/drivers/shared_all/src/initialize_veg.c
```diff
@@ -39,31 +39,33 @@ initialize_veg(veg_var_struct **veg_var,
...
+            veg_var[i][j].albedo = 0.0;
+            veg_var[i][j].displacement = 0.0;
+            veg_var[i][j].fcanopy = 0.0;
             veg_var[i][j].LAI = 0.0;
+            veg_var[i][j].roughness = 0.0;
+            veg_var[i][j].throughfall = 0.0;
+            veg_var[i][j].Wdew = 0.0;
             veg_var[i][j].Wdmax = 0.0;
-            veg_var[i][j].fcanopy = 0.0;
-            veg_var[i][j].albedo = 0.0;
```
yixinmao: Should `veg_var[i][j].canopyevap` be set to zero also?

vic/drivers/shared_all/src/initialize_veg.c
```diff
@@ -39,31 +39,33 @@ initialize_veg(veg_var_struct **veg_var,
...
         }
         if (options.CARBON) {
             for (j = 0; j < options.SNOW_BAND; j++) {
```
yixinmao: For the carbon part, also `.Litterfall` and `.NPPfactor` ?

vic/drivers/shared_all/src/initialize_veg.c
tbohn: It seems like it should.  However, I would suggest that we create a separate issue for that - an issue that would list all the things we need to change to make restarts be exact.  This PR is only for the purpose of allowing us to remove instances of `month` from vic_run().

vic/drivers/shared_all/src/initialize_veg.c
tbohn: Again, that should be part of an issue for restarts.  This PR's purpose is only to remove instances of month from vic_run().  Anything else is outside the scope of this PR.

PR #481, opened 2016-04-24
Clean up and initial fix for exact restart - classic driver

yixinmao:
#### Changes related to exact restart:
* Raised ascii state variable outputs to %.16g
* Added the prognostic state var `energy.Tfoliage` to write and read in the state file

#### Clean up:
* Cleanly separated state and flux variables in vic_def.h and print_library_shared.c
* In `initialize_*.c`, all state and flux variables are now initialized to zero (before some were missing); in `initalize_soil.c`, moved computation for derived variables into `compute_derived_state_vars`.
* In `compute_derived_state_vars`, only computed the four variables when `FROZEN_SOIL=TRUE`:

    energy.Cs_node
    energy.kappa_node
    energy.moist
    energy.ice


vic/drivers/image/Makefile
```diff
@@ -156,4 +156,4 @@ depend: .depend
 	$(CC) $(CFLAGS) -M $(SRCS) > $@
 
 clean::
-	\rm -f .depend
+	\rm -f .depend
```
jhamman: Can you revert the changes to the image driver Makefile?  They are not part of this PR and I don't think we want to hard code the hydra paths in there like this.

vic/drivers/shared_all/src/initialize_soil.c
```diff
@@ -38,28 +38,40 @@ initialize_soil(cell_data_struct **cell,
...
+            cell[veg][band].CInter = 0.0;
+            cell[veg][band].CSlow = 0.0;
             for (lindex = 0; lindex < options.Nlayer; lindex++) {
-                cell[veg][band].layer[lindex].evap = 0;
-                cell[veg][band].layer[lindex].moist = 0;
-                tmp_moist[lindex] = cell[veg][band].layer[lindex].moist;
+                cell[veg][band].layer[lindex].Cs = 0.0;
+                cell[veg][band].layer[lindex].T = 0.0;
+                cell[veg][band].layer[lindex].bare_evap_frac = 0.0;
+                cell[veg][band].layer[lindex].evap = 0.0;
```
jhamman: isn't this a flux?

vic/drivers/shared_all/src/initialize_soil.c
```diff
@@ -38,28 +38,40 @@ initialize_soil(cell_data_struct **cell,
...
-                    cell[veg][band].layer[lindex].ice[frost_area] = 0;
+                    cell[veg][band].layer[lindex].ice[frost_area] = 0.0;
                 }
+                cell[veg][band].layer[lindex].kappa = 0.0;
+                cell[veg][band].layer[lindex].moist = 0.0;
+                cell[veg][band].layer[lindex].phi = 0.0;
             }
-            compute_runoff_and_asat(soil_con, tmp_moist, 0,
-                                    &(cell[veg][band].asat), &tmp_runoff);
-            wrap_compute_zwt(soil_con, &(cell[veg][band]));
```
jhamman: do we need to initialize `cell[veg][band].asat`?

vic/drivers/shared_all/src/print_library_shared.c
```diff
@@ -438,18 +446,18 @@ print_layer_data(layer_data_struct *ldata,
     size_t i;
 
     fprintf(LOG_DEST, "layer_data:\n");
-    fprintf(LOG_DEST, "\tCs   : %.4f\n", ldata->Cs);
-    fprintf(LOG_DEST, "\tT    : %.4f\n", ldata->T);
-    fprintf(LOG_DEST, "\tevap : %.4f\n", ldata->evap);
+    fprintf(LOG_DEST, "\tCs   : %f\n", ldata->Cs);
+    fprintf(LOG_DEST, "\tT    : %f\n", ldata->T);
+    fprintf(LOG_DEST, "\tevap (flux): %f\n", ldata->evap);
```
jhamman: in what case is `evap` not a flux?

vic/vic_run/src/surface_fluxes.c
```diff
@@ -250,7 +250,7 @@ surface_fluxes(bool                 overstory,
 
     energy->advection = 0;
     energy->deltaCC = 0;
-    if (snow->swq > 0) {
+
         snow_flux = energy->snow_flux;
     }
     else {
```
jhamman: this change appears to be inadvertent or incomplete.

jhamman:
@yixinmao - I'm made a few minor comments.  You have failing tests on all drivers and there are a bunch of compiler warnings to address. Once you've cleaned thigs up here, ping @tbohn for the next review.

vic/drivers/classic/src/write_model_state.c
```diff
@@ -146,7 +146,7 @@ write_model_state(all_vars_struct *all_vars,
                    filep->statefile);
         }
         else {
-            fprintf(filep->statefile, " %f ", soil_con->dz_node[nidx]);
+            fprintf(filep->statefile, " %.16g ", soil_con->dz_node[nidx]);
```
jhamman: suggest defining a header variable to handle this repetitive format. E.g.

```C
#define ASCII_STATE_FLOAT_FMT " %.16g "
```

Or something with a shorter but descriptive name?

yixinmao: Sure will do.

vic/drivers/image/Makefile
yixinmao: Yes this should not be committed...

vic/drivers/shared_all/src/initialize_soil.c
yixinmao: Yes you are right. Will separate it (and also `bare_evap_frac`)

vic/drivers/shared_all/src/initialize_soil.c
yixinmao: why not?

vic/drivers/shared_all/src/print_library_shared.c
yixinmao: Hmm I added this because `evap` is actually a flux, but it got mixed in the state section in the layer structure. Will reorder the structure to make it clearer.

tbohn: @yixinmao did you notice that there already was a line for evap?  So you created a duplicate line...

yixinmao: @tbohn where is the duplicated line? (I'm reordering this section so that `layer.evap` and `layer.bare_evap_frac` will be separated from the other state variables)

tbohn: Oh, sorry, I mis-read it.  Never mind!

vic/vic_run/src/surface_fluxes.c
yixinmao: Yes, I somehow deleted this line accidentally without re-testing the run. My fault, will fix it.

vic/drivers/shared_all/src/initialize_soil.c
yixinmao: Actually did you mean you think we should initialize `cell[veg][band].asat`? This part is done in `compute_derived_state_vars`

yixinmao:
@jhamman I should have addressed all your comments (thank you!). @tbohn do you want to take a look at the current version again?

vic/drivers/classic/src/write_model_state.c
```diff
@@ -293,10 +298,19 @@ write_model_state(all_vars_struct *all_vars,
                            filep->statefile);
                 }
                 else {
-                    fprintf(filep->statefile, " %f", energy[veg][band].T[nidx]);
+                    fprintf(filep->statefile, " "ASCII_STATE_FLOAT_FMT, energy[veg][band].T[nidx]);
                 }
             }
 
+            /* Write foliage temperature */
```
tbohn: Need to also increment the Nbytes count to include the size of Tfoliage, on line 103 (I can't seem to add an inline comment to that line of this diff, so I'm adding the note here).  You can look at how the other variables are done and pattern Tfoliage after them.  It should be a simple 1-liner.

vic/drivers/shared_all/src/initialize_veg.c
```diff
@@ -39,35 +39,44 @@ initialize_veg(veg_var_struct **veg_var,
...
         for (j = 0; j < options.SNOW_BAND; j++) {
+            // Prognostic states
             veg_var[i][j].albedo = 0.0;
             veg_var[i][j].displacement = 0.0;
             veg_var[i][j].fcanopy = 0.0;
             veg_var[i][j].LAI = 0.0;
             veg_var[i][j].roughness = 0.0;
-            veg_var[i][j].throughfall = 0.0;
             veg_var[i][j].Wdew = 0.0;
             veg_var[i][j].Wdmax = 0.0;
```
tbohn: Technically, Wdmax is a derived (or "diagnostic", I suppose) state, completely derived from LAI.  However, both LAI and Wdmax get updated at similar times (every time that LAI gets a new value, Wdmax is recomputed), so keeping the initialization of Wdmax here would be completely consistent with that other behavior.  It's your (and/or Bart and Joe's) call.

tbohn:
@yixinmao I've finished looking it over.  Other than my 2 comments, it looks good.  This is of course an important improvement in making the restarts more accurate.  It also helps clean up the code, which will help us maintain the code in the future.

vic/drivers/classic/src/write_model_state.c
yixinmao: Good point! Will do. I haven't really tested for binary format - should we actually have a binary version of the sample data?

tbohn: Do you mean a binary version of the initial state file?  (whether the state files are binary is independent of whether the forcings or outputs are binary)

yixinmao: I see. I was thinking about a complete binary sample dataset, but specifically to this problem testing binary i/o for the state file is enough. Thanks!

vic/drivers/shared_all/src/initialize_veg.c
yixinmao: Hmm what exactly do you mean and what do you suggest we should do?

tbohn: Personally I'm OK with leaving it as it currently is.  I was just pointing out that Wdmax is derived from LAI - it's good to know this if we want to avoid future problems similar to the one you just fixed (for example, someone who doesn't know it's derived from LAI might change LAI somewhere else in the code and forget to include an update to Wdmax).  I don't know if there is a different way to handle updating Wdmax that is not also cumbersome (since LAI is updated in a few different places).  I suppose one way to handle it could be to eliminate Wdmax from veg_var, veg_con, and veg_lib structures, and instead rewrite the logic that uses Wdmax to compute Wdmax from LAI on the spot immediately before it is used, and for that particular Wdmax to be a temporary variable that exists only within that function.  But that might be too much trouble. @jhamman and @bartnijssen might want to weight in on this.

jhamman: I agree that `Wdmax` isn't really a state or flux variable, it is really a derived model parameter. my suggestion would be to open a new issue that rethinks the handling of `Wdmax` and we can tackle that in another PR sometime down the road.

jhamman: xref #488

PR #485, opened 2016-04-27
Add new meteorological forcing data type, sunshine duration (h).

Sibada:
In our research we have not short wave or cloud cover data but we have sunshine duration data with a so-so quality during 1950-2014. And we find that potential evapotranspiration data estimated in Pearl River Basin, China is rather inaccurate (it increased significantly, but other exam showed that it should be decreased obviously, while sunshine duration decreased during those years). So we manage to add sunshine duration into our model.


jhamman:
@Sibada - 

Thank you for issuing a pull request.  However, I don't think we will be able to accept this feature. The MTCLIM meteorologic forcing pre-processor has been depreciated in VIC 5 (#288)  which we will be releasing shortly.  In the future, we hope you will continue to make contributions to VIC.  Next time please inquire with us ahead of time to check on their applicability.

Issue #463, opened 2016-04-08
Memory leaks


tbohn:
Using valgrind to run the latest develop branch image driver on Stehekin test case, found unfreed memory in the following places:

Image Driver:
vic_init.c: 74
alloc_atmos.c: 41, 45
vic_mpi_support.c: 61, 62, 63
get_nc_field.c: 44, 124
get_nc_dimension.c: 42
make_veg_var.c: 46
make_snow_data.c: 50
make_cell_data.c: 44
make_energy_bal.c: 49
vic_finalize.c: 106  (from @jhamman's comment below)

Classic Driver:
set_output_defaults.c: 94

Note: some of the line numbers may change as commits continue to be merged into the code.

jhamman:
I'll also add: https://github.com/UW-Hydro/VIC/pull/460/files#r58979183
vic_finalize.c: 106

tbohn:
I'll work on this today & over the weekend.

tbohn:
I've been able to solve the following leaks:
vic_init.c
alloc_atmos.c
make_veg_var.c
make_cell_data.c
make_snow_data.c
make_energy_bal.c
set_output_defaults.c

The remaining leaks are related to creation of MPI data structs (mpi_nc_file_struct_type, mpi_option_struct_type, and mpi_param_struct_type) and the opening of nc files in get_nc_dimension.c and get_nc_field.c.

For the MPI data structs, there is a call to MPI_Finalize(), which one would think would free all MPI-related data structures, in vic_image.c.  So, not sure what to do there.  @jhamman suggested possibly the function MPI_Type_free().

For the nc_file opening in get_nc* functions, I'm not sure why these are happening, since nc_close() is called at the end of each function.

tbohn:
@jhamman There still are 3 outstanding leaks in get_nc_dimensions.c and get_nc_field.c that I wasn't able to fix in my PR (#490).  So it might be premature to close this issue...  Or we should open a new issue for just those leaks...

bartnijssen:
Do we still have memory leaks? Would be good to list the remaining ones in some detail. Ideally it would state which version of the driver, the platform (OS and version), nc library (version) and compiler (name and version).

tbohn:
Yes, we still have the following memory leaks for image driver, os = gnu/linux x86_64, compiler = gcc 4.4.7, netcdf 4 (I think; it's whatever is on the UW lab machines):
get_nc_field.c: 44, 124
get_nc_dimension.c: 42
[...]

PR #490, opened 2016-05-03
Fixed memory leaks in some functions.

tbohn:
This PR fixes the following memory leaks in VIC 5.0 (from issue #463 ):

Image Driver:
```shell
vic_init.c: 74
alloc_atmos.c: 41, 45
make_veg_var.c: 46
make_snow_data.c: 50
make_cell_data.c: 44
make_energy_bal.c: 49
```

Classic Driver:
```shell
set_output_defaults.c: 94
```

This does not (yet) completely resolve issue #463 but gets us part-way there, anyway.  If preferable, I can keep adding fixes to this PR until they're all fixed.

### DIFF
--- a/vic/drivers/classic/include/vic_driver_classic.h
+++ b/vic/drivers/classic/include/vic_driver_classic.h
@@ -48,6 +48,7 @@ void compute_cell_area(soil_con_struct *);
 size_t count_outfile_nvars(FILE *gp);
 out_data_struct *create_output_list();
 void free_atmos(int nrecs, atmos_data_struct **atmos);
+void free_out_data_files(out_data_file_struct **);
 void free_veg_hist(int nrecs, int nveg, veg_hist_struct ***veg_hist);
 void free_veglib(veg_lib_struct **);
 double get_dist(double lat1, double long1, double lat2, double long2);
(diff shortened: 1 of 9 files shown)


tbohn:
Now I've fixed the issues with the mpi data structures, thanks to @jhamman 's suggestion of MPI_Type_free().

jhamman:
@tbohn - thanks for doing this.  I'm glad `MPI_Type_free()` was what you were looking for.  Can you explain for the record how you found these memory errors/leaks?  I'm wondering if it would make since to   add a check for memory leaks in a few of our travis tests.

tbohn:
Sure: `valgrind -v --leak-check=full /path_to_vic/vic_image.exe -g global_file >& log_file`

The log file will contain an error summary at the end that lists the occurrences of leaks.  It will also summarize occurrences of using uninitialized variables.

jhamman:
Okay, cool. I'll open a separate issue for the CI testing of memory leaks. We'll just need to add the `--error-exitcode` flag to your command above:

`valgrind -v --leak-check=full --error-exitcode=-1 vic_image.exe -g global_file`

> --error-exitcode=<number> [default: 0]
Specifies an alternative exit code to return if Valgrind reported any errors in the run. When set to the default value (zero), the return value from Valgrind will always be the return value of the process being simulated. When set to a nonzero value, that value is returned instead, if Valgrind detects any errors. This is useful for using Valgrind as part of an automated test suite, since it makes it easy to detect test cases for which Valgrind has reported errors, just by inspecting return codes.

xref: #492

PR #491, opened 2016-05-03
changes to system test infrastructure and added tests to check output a…

dgergel:
…nd domain files for image driver


tests/run_tests.py
```diff
@@ -9,13 +9,17 @@
 import datetime
 from collections import OrderedDict
 import string
+import numpy as np
+import xarray as xr
```
jhamman: can you add `xarray` and `bottleneck` to `ci/requirements.yml`?

tests/run_tests.py
```diff
@@ -292,7 +296,7 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
 
         # ------------------------------------------------------------ #
         # read template global parameter file
-        infile = os.path.join(test_dir, 'samples',
+        infile = os.path.join(test_data_dir + 'image/Stehekin/parameters/',
```
jhamman: I don't think we want this change.

tests/run_tests.py
```diff
@@ -339,6 +343,8 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
         test_comment = ''
         error_message = ''
 
+        print(vic_exe) 
+
```
jhamman: remove print statement.

tests/run_tests.py
```diff
@@ -351,28 +357,37 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
+                            if (start is not None) and (end is not None):
+                                check_completed(df, start, end)
+                            else:
+                                start = df.index[0]
+                                end = df.index[-1]
+            elif driver == "image": 
+                if test_dict['check']:
                     # check for nans in the df
                     if 'nonans' in test_dict['check']:
                         check_for_nans(df)
```
jhamman: this `check_for_nans()` belongs in the classic driver section above.

tests/run_tests.py
```diff
@@ -351,28 +357,37 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
-            # -------------------------------------------------------- #
+            
+                    # check for nans in domain file  
+                    if 'output_nans' in test_dict['check']:
+                        print(test_dict['domain_file'])
```
jhamman: remove print statement

tests/run_tests.py
```diff
@@ -351,28 +357,37 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
+                        ds_domain = xr.open_dataset(os.path.join(test_data_dir + 'image/Stehekin/parameters/',
+                                                    test_dict['domain_file']))
+                        ds_output = xr.open_dataset(os.path.join(dirs['results'], 'Stehekin.history.nc'))
```
jhamman: We'll want to remove these hard coded paths and get everything from the test config file.

tests/run_tests.py
```diff
@@ -767,6 +782,8 @@ def replace_global_values(gp, replace):
     gpl = []
     for line in iter(gp.splitlines()):
         line_list = line.split()
+        if line_list == []:
+            continue
```
jhamman: is this just skipping empty lines? If so, it would be cleaner to write it as:

```Python
if not line_list:
    continue
```

it may be better to go without the continue statement and just do:

```Python
if line_list:
    ... # complete the rest of the for loop
```

tests/system/system_tests.cfg
```diff
@@ -1,395 +1,451 @@
...
+#
+#[System-run_baseflow_nijssen]
+#driver = classic
+#test_description = complete simulation if BASEFLOW=NIJSSEN2001
+#global_parameter_file = global_param.ROSEM.system.txt
+#expected_retval = 0
+#check = None
+#[[options]]
+#BASEFLOW=NIJSSEN2001
+#'''
```
jhamman: I think we should just remove all these tests until you or @yixinmao can fix the classic driver test system.

tests/system/system_tests.cfg
```diff
@@ -1,395 +1,451 @@
...
+
+[System-check-domainfile-and-nans] 
+driver = image
+test_description = check that NaNs in domain file and output file match
```
jhamman: check that the coordinates and mask in the domain and output files match.

tests/run_tests.py
```diff
@@ -351,28 +357,37 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
-                    start = None
-                    end = None
-
-                for fname in glob.glob(os.path.join(dirs['results'], '*')):
-                    df = read_vic_ascii(fname, header=True)
-                    # do numeric tests
-
-                    # check that each dataframe includes all timestamps
+            # check output files (different tests depending on driver)
+            if driver == "classic":
```
yixinmao: Would it be better to reorder the if statements, say, put `if test_dict['check']` as the highest level, and then check for each possible checks, and then the driver options if needed? Something like this:
```
if test_dict['check']:
        if 'complete' in test_dict['check']:
                do something (maybe another if statement for classic or image driver, if needed)
        if 'nonan' in test_dict['check']:
                do something
```

To me it would be cleaner, and another reason I'm suggesting this is because for some more complicated tests (e.g., the restart tests), maybe only the reading data part is different for different drivers, but the other checking process would be the same. In those cases, it makes more sense to put the distinguishing of drivers in the most inner layer, so that the same checking function can be called.

tests/run_tests.py
yixinmao: Haha sorry I think actually I originally added these two lines... and yes they are for skipping empty lines (otherwise the function breaks at empty lines). Good suggestion.

tests/system/system_tests.cfg
```diff
@@ -1,395 +1,451 @@
...
+#BASEFLOW=NIJSSEN2001
+#'''
+
+[System-check-domainfile-and-nans] 
+driver = image
+test_description = check that NaNs in domain file and output file match 
+global_parameter_file = global_param.image.STEHE.feb.txt 
+domain_file = domain.stehekin.20151028.nc
+expected_retval = 0
+check = output_nans
```
yixinmao: To me the check names `nonans` and `output_nans` are kind of confusing. Maybe change for some more self-explaining names?

tests/system/system_tests.cfg
```diff
@@ -1,395 +1,451 @@
...
+#expected_retval = 0
+#check = None
+#[[options]]
+#BASEFLOW=NIJSSEN2001
+#'''
+
+[System-check-domainfile-and-nans] 
+driver = image
+test_description = check that NaNs in domain file and output file match 
+global_parameter_file = global_param.image.STEHE.feb.txt
```
jhamman: Are you using the global param file in the `examples` directory? We probably need more path info here too.

tests/system/system_tests.cfg
```diff
@@ -1,395 +1,451 @@
...
+#check = None
+#[[options]]
+#BASEFLOW=NIJSSEN2001
+#'''
+
+[System-check-domainfile-and-nans] 
+driver = image
+test_description = check that NaNs in domain file and output file match 
+global_parameter_file = global_param.image.STEHE.feb.txt 
+domain_file = domain.stehekin.20151028.nc
```
jhamman: I think we'll need more path info here.

tests/system/system_tests.cfg
```diff
@@ -1,395 +1,451 @@
...
+expected_retval = 0
+check = output_nans
+[[options]]
```
jhamman: Do we need this options section here?  If not, remove it.

tests/test_image_driver.py
```diff
@@ -0,0 +1,64 @@
...
+
+import xarray as xr
+import numpy as np 
+import numpy.testing as npt 
+
+
+# In[28]:
+
+file = '/Users/diana/workdir/VIC_tests_20160331/examples/Example-Image-Stehekin-base-case/results/Stehekin.history.nc'
+domainfile = '/Users/diana/Dropbox/UW/VIC_sample_data/image/Stehekin/parameters/domain.stehekin.20151028.nc'
```
jhamman: remove hard coded paths.

tests/test_image_driver.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+# In[29]:
+
+ds = xr.open_dataset(file)
+ds_domain = xr.open_dataset(domainfile)
```
jhamman: remove these lines as we don't want to run them every time we import this module.

tests/test_image_driver.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    
+    # extract mask from domain file 
+    mask = ds_domain['mask']
+    
+    # check that lats and lons are the same 
+    try: 
+        npt.assert_array_equal(ds.lat, ds_domain.lat) and npt.assert_array_equal(ds.lon, ds_domain.lon)
+    except AssertionError: 
+        print("Domains for domain file and output files are different")
```
yixinmao: Here we actually do want to raise an error?

tests/test_image_driver.py
```diff
@@ -0,0 +1,64 @@
...
+            try:
+                npt.assert_array_equal(np.isnan(da_sel.values), np.isnan(mask.values))
+            except AssertionError:
+                print("NaN values in output file and domain file do not match up in %s" % da)
+
+
+# In[59]:
+
+# calling function 
+assert_nan_equal(ds_domain, ds)
```
yixinmao: What is this line for?

tests/run_tests.py
```diff
@@ -9,13 +9,17 @@
...
+import xarray as xr
+import numpy.testing as npt 
 
 import pytest
 import pandas as pd
 
 from tonic.models.vic.vic import VIC, VICRuntimeError  # , read_vic_ascii
 from tonic.io import read_config, read_configobj
 from tonic.testing import check_completed, check_for_nans, VICTestError
+from test_image_driver import assert_nan_equal
```
yixinmao: Hmm I think this way of calling functions from another file only works when you run the `run_tests.py` script from the directory where the scripts are. @jhamman do we want to put some functions in a separate file, and then call it in some more general way (how?)? Or do we want to just put all the functions in `run_tests.py`?

tests/test_image_driver.py
jhamman: Let's rework this a bit:

```Python
def assert_nan_equal(ds_domain, ds_output, match_vars=['mask', 'lat', 'lon']):
    ...
    for var in match_vars:
        npt.assert_all_close(ds[var], ds_domain[var], equal_nan=True)
        # We want an AssertionError to be raised if they don't match.

    ....
```

tests/system/system_tests.cfg
yixinmao: Can the path of the domain file be read from the global parameter file?

tests/test_image_driver.py
```diff
@@ -0,0 +1,64 @@
...
+
```
jhamman: remove these lines.

tests/run_tests.py
yixinmao: If we assume that the global parameter template file is under a certain directory, it should probably be put under `system` subdirectory under `test_dir`, so `infile = os.path.join(test_dir, 'system', test_dict['global_parameter_file'])`

tests/test_image_driver.py
```diff
@@ -0,0 +1,64 @@
...
+    # check that lats and lons are the same 
+    try: 
+        npt.assert_array_equal(ds.lat, ds_domain.lat) and npt.assert_array_equal(ds.lon, ds_domain.lon)
+    except AssertionError: 
+        print("Domains for domain file and output files are different")
+    
+    for da in ds_output.data_vars: 
+    
+        # slice first time step of output DataArray 
+        da_sel = ds_output[da].isel(time=0)
```
jhamman: we want to check all timesteps.

tests/system/system_tests.cfg
yixinmao: See my comments above, one reasonable way is to assume the global param template file is in the `test_dir/system`. Although there may be other more generic way to handle the path.

tests/test_image_driver.py
```diff
@@ -0,0 +1,64 @@
...
+                    npt.assert_array_equal(np.isnan(da_sel.isel(nlayer=layer).values), np.isnan(mask.values))
+                except AssertionError:
+                    print("NaN values in output files and domain files do not match up in %s layer %1f" % (da, layer))
+                    
+        else: 
+            # check that nans are occurring in the same place in the two arrays 
+            try:
+                npt.assert_array_equal(np.isnan(da_sel.values), np.isnan(mask.values))
+            except AssertionError:
+                print("NaN values in output file and domain file do not match up in %s" % da)
```
jhamman: This section will need be generalized further. You can assume that the spatial dimension names in the two datasets are the same (e.g. lat/lon) but you won't know what they are (so get them from the mask in the domain file).

I would use `pd.isnull` on the full arrays then compare 2d slices after summing along the non-spatial dims.

Again, if they don't match, we need to raise the `AssertionError`.

tests/system/system_tests.cfg
```diff
@@ -1,395 +1,451 @@
...
+#test_description = complete simulation if BASEFLOW=NIJSSEN2001
+#global_parameter_file = global_param.ROSEM.system.txt
+#expected_retval = 0
+#check = None
+#[[options]]
+#BASEFLOW=NIJSSEN2001
+#'''
+
+[System-check-domainfile-and-nans] 
+driver = image
```
yixinmao: This is more like an issue for later discussion, not necessarily for this PR I guess - so based on our conversation yesterday, I agree with you that it would be cleaner to put the driver option here for each individual system test. But currently driver as well as VIC executable file are still input arguments to `run_tests.py`. I feel this is a little confusing with the driver specification here, but I guess for unit tests we do want to specify the driver from input command line (right?). So maybe we can talk how we want to handle this eventually...

tests/run_tests.py
dgergel: Sorry about these print statements! I forgot to remove them earlier.

tests/system/system_tests.cfg
dgergel: @jhamman even if we may want to salvage some of them?

tests/run_tests.py
dgergel: I like this idea. I've standardized the test name to 'output_nans' for both drivers.

tests/system/system_tests.cfg
dgergel: adding more informative name :-)

tests/system/system_tests.cfg
dgergel: I've gone ahead and used @yixinmao's suggestion to use the system subdir in `test_dir`. And yes, it's the global param file from the examples directory, not the old one.

tests/system/system_tests.cfg
dgergel: I've put this into the same system subdir in `test_dir` for now.

tests/system/system_tests.cfg
jhamman: Yes. That is why we have version control.  We try not to commit commented out code.

dgergel:
@jhamman and @yixinmao I've addressed your comments. The test function may need a bit more work but I wanted to go ahead so that @yixinmao can add restart tests while I'm in the field tomorrow.

tests/system/system_tests.cfg
```diff
@@ -1,395 +1,7 @@
-[System-base]
-test_description = Short simulation to ensure system tests configuration works
-global_parameter_file = global_param.ROSEM.system.txt
+[System-check-domainfile-and-nans] 
+driver = image
+test_description = check that coordinates and mask in domain and output files match 
+global_parameter_file = global_param.image.STEHE.feb.txt
```
yixinmao: Hmm I think we want to commit this global parameter template file as well? (also a minor point - I personally think no need to put "feb" in the file name, since this global file should work both for water balance and energy balance I suppose?)

yixinmao:
You didn't really change the VIC code, but it seems that the travis tests have failed. Is it possibly because at the beginning of `run_tests.py` it tries to import the other file you added, but it doesn't work unless it's been running under the script directory?

tests/run_tests.py
```diff
@@ -351,28 +355,38 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
+                        if driver == "classic": 
+                            # check for nans in the df 
+                            check_for_nans(df) 
+
+                        elif driver == "image": 
+                            # check for nans in domain file  
+                            ds_domain = xr.open_dataset(os.path.join(test_dir, 'system', 
+                                                        test_dict['domain_file']))
+                            ds_output = xr.open_dataset(os.path.join(dirs['results'], 
+                                                        'Stehekin.history.nc'))
```
jhamman: Instead of hard-coding this file name, can we do a `glob` wildcard (`*.nc`) inside the results directory and check all files present?

tests/test_image_driver.py
```diff
@@ -0,0 +1,44 @@
...
+        for ts in range(0, len(ds_output.time)):
+     
+            # slice time step of output DataArray
+            da_sel = ds_output[da].isel(time=ts) 
+
+            # check all layers 
+            if (len(ds_output[da].dims) > 3):  
+                for layer in range(0, len(ds_output[da].nlayer)):
+                    # npt.assert_allclose(da_sel.isel(nlayer=layer), ds_domain['mask'], equal_nan=True)
+                    npt.assert_array_equal(np.isnan(da_sel.isel(nlayer=layer).values), np.isnan(ds_domain['mask'].values))
```
jhamman: break line at 80 chars (PEP8).

tests/test_image_driver.py
```diff
@@ -0,0 +1,44 @@
...
+            da_sel = ds_output[da].isel(time=ts) 
+
+            # check all layers 
+            if (len(ds_output[da].dims) > 3):  
+                for layer in range(0, len(ds_output[da].nlayer)):
+                    # npt.assert_allclose(da_sel.isel(nlayer=layer), ds_domain['mask'], equal_nan=True)
+                    npt.assert_array_equal(np.isnan(da_sel.isel(nlayer=layer).values), np.isnan(ds_domain['mask'].values))                    
+                    
+            else: 
+                # npt.assert_allclose(da_sel, ds_domain['mask'], equal_nan=True)
```
jhamman: remove commented out code

tests/test_image_driver.py
```diff
@@ -0,0 +1,44 @@
...
+        # check all timesteps 
+        for ts in range(0, len(ds_output.time)):
+     
+            # slice time step of output DataArray
+            da_sel = ds_output[da].isel(time=ts) 
+
+            # check all layers 
+            if (len(ds_output[da].dims) > 3):  
+                for layer in range(0, len(ds_output[da].nlayer)):
+                    # npt.assert_allclose(da_sel.isel(nlayer=layer), ds_domain['mask'], equal_nan=True)
```
jhamman: remove commented out code

tests/test_image_driver.py
```diff
@@ -0,0 +1,44 @@
...
+    """ 
+    test to see if nans in two data arrays are the same in order to check domain 
+    first dataarray is domain file, second dataarray is a variable in the output
+    dataset for image driver 
+    """
+
+    # get lat/lng dim names from domain file 
+    lng_name, lat_name = ds_domain.coords
+
+    match_vars = [lng_name, lat_name]
```
jhamman: `coords` will not always unpack in the order you are asserting it will here. It actually doesn't matter so you can generalize this.

tests/test_image_driver.py
```diff
@@ -0,0 +1,44 @@
...
+        # raise AssertionError if they don't match
+        npt.assert_allclose(ds_output[var], ds_domain[var], equal_nan=True) 
+
+    # check that nans are occurring in the same place in the arrays 
+    
+    # check all variables in the dataset 
+    for da in ds_output.data_vars: 
+        
+        # check all timesteps 
+        for ts in range(0, len(ds_output.time)):
```
jhamman: Nothing outright wrong here but the for loop isn't required. I still think using `isull` makes more since here. Some pseudo code for you:

```Python
# Find grid cells that are never nans (that mask should match the domain mask)
da_notnull = da.notnull().all(dim=('time', ..., other_non_space_dims))
# compare the masks
npt.assert_array_equal(da.values, ds_domain['mask'].values)
```

dgergel:
@jhamman this should be ready to merge.

tests/test_image_driver.py
```diff
@@ -0,0 +1,39 @@
...
+
+def assert_nan_equal(ds_domain, ds_output): 
+    """ 
+    test to see if nans in two data arrays are the same in order to check domain 
+    first dataarray is domain file, second dataarray is a variable in the output
+    dataset for image driver 
+    """
+
+    # get lat/lng dim names from domain file 
+    match_vars = ds_domain.coords
```
jhamman: You actually get more than the `dim` names here, you get the coordinates. You should add this distinction to your comment.

I would also change this line slightly to make sure we only check the coordinates that are in the `mask` variable:

`match_vars = ds_domain['mask'].coords`

tests/test_image_driver.py
```diff
@@ -0,0 +1,39 @@
...
+            # check all layers and timesteps 
+            da_isnull_reduced = ds_output[da].isnull().all(dim=('time', 'nlayer')) 
+            npt.assert_array_equal(da_isnull_reduced.values, 
+                                    np.isnan(ds_domain['mask']))
+
+        else: 
+            # check all timesteps 
+            da_isnull_reduced = ds_output[da].isnull().all(dim=('time')) 
+            npt.assert_array_equal(da_isnull_reduced.values, 
+                                    np.isnan(ds_domain['mask']))
```
jhamman: We shouldn't need to specify the `nlayer` dim option or even check if the variable is 4d.

What you want to do is take the difference in the dims between what the `mask` variable has and what your data variable has.  For example:

```Python
In [1]: ds_domain.mask.dims
Out[1]: ('x', 'y')
In [2]: ds_output[da].dims
Out[2]: ('x', 'y', 'time', 'snow_band', 'nlayer')
```

You can do this by taking the symmetric difference:

```Python
In [3]: a = ('x', 'y')
In [4]: b = ('x', 'y', 'time', 'snow_band', 'nlayer')
In [5]: set(a) ^ set(b)
Out[5]: {'nlayer', 'snow_band', 'time'}
```

Then you want to do you `...all(dim=(all_diff_dims)`.  Basically you are collapsing every data variable down to the shape of the mask.

tests/test_image_driver.py
dgergel: updated.

tests/test_image_driver.py
dgergel: So cool! Updated.

jhamman:
**Good news** - Everything's passing and this looks ready to go.
**Bad news** - you aren't actually testing anything with this yet.

Let's do two things here:
- use this new tool to test the example output, I think you'll just need to edit this line: https://github.com/UW-Hydro/VIC/blob/develop/tests/examples/examples.cfg#L29
- since you've cleaned out the system tests, let's also run those on travis: you should change this line in both the image and classic drivers: https://github.com/UW-Hydro/VIC/blob/develop/ci/classic.travis#L61

dgergel:
@jhamman I've updated the `run_examples` function in `run_tests.py` to include NaN checks (and restructured it so that it matches the `run_system' function for classic/image driver tests).

tests/run_tests.py
```diff
@@ -602,7 +616,7 @@ def run_examples(config_file, vic_exe, test_data_dir, out_dir, driver):
 
         # ------------------------------------------------------------ #
         # read template global parameter file
-        infile = os.path.join(test_dir, 'examples', test_dict['global_parameter_file'])
+        infile = os.path.join(test_dir, 'examples', test_dict['global_parameter_file'])
```
jhamman: remove extra whitespace.

tests/run_tests.py
```diff
@@ -292,7 +296,7 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
 
         # ------------------------------------------------------------ #
         # read template global parameter file
-        infile = os.path.join(test_dir, 'samples',
+        infile = os.path.join(test_dir, 'system',
```
jhamman: remove extra whitespace

PR #496, opened 2016-05-06
Fixed forcing unit conversion bug in image driver

yixinmao:
- Keeping the air pressure in Pa (as it is read in)
    - Also fixed some comments related to unit description

### DIFF
--- a/vic/drivers/image/src/vic_force.c
+++ b/vic/drivers/image/src/vic_force.c
@@ -334,13 +334,12 @@ vic_force(void)
             atmos[i].air_temp[j] -= CONST_TKFRZ;
             // precipitation in mm/period
             atmos[i].prec[j] *= global_param.snow_dt;
-            // pressure in kPa
-            atmos[i].pressure[j] /= PA_PER_KPA;
-            // vapor pressure in kPa (we read specific humidity in kg/kg)
+            // pressure in Pa
+            // vapor pressure in Pa (we read specific humidity in kg/kg)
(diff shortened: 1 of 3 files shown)


vic/drivers/image/src/vic_force.c
```diff
@@ -334,13 +334,12 @@ vic_force(void)
             atmos[i].air_temp[j] -= CONST_TKFRZ;
             // precipitation in mm/period
             atmos[i].prec[j] *= global_param.snow_dt;
-            // pressure in kPa
-            atmos[i].pressure[j] /= PA_PER_KPA;
```
jhamman: @wietsefranssen - is this the change you had mentioned a few weeks back?

yixinmao:
@jhamman do we want to merge this in?

vic/drivers/image/src/vic_force.c
wietsefranssen: @yixinmao: So the units of air pressure (input and output) should be Pa? Good to know! This was still not clear for me :).

But still: with this fix the output is still a factor 1000 smaller than input/forcing (without the fix it is even a factor 1.000.000 smaller).

yixinmao: @wietsefranssen : currently the input units for classic driver and image driver are different - for example, for air pressure, the classic driver input is in kPa, and the image driver input is in Pa. As for output - the classic driver output pressure should be in kPa; I haven't looked in detail at the image driver output units yet, and it is possible that there might be a bug when outputing data. We are looking at the output part of the image driver soon, and will pin you if we make any changes related to the output units.

PR #498, opened 2016-05-10
Fixed call to water_balance() in vic_run().

### LINKED ISSUE #497: Potential typo in vic_run() call to water_balance()
Currently, line 478 of vic_run.c calls water_balance() (part of lake model) and one of the arguments in the call is *veg_con.  Within vic_run(), veg_con is an array of structures, one per veg tile, so *veg_con doesn't quite make sense (I think it evaluates to veg_con[0]).  But water_balance() is expecting the veg_con structure that corresponds to the veg_tile that actually contains the lake, which might not be veg_con[0].  So it looks like a typo.

The fix would be simple: replace *veg_con with veg_con[iveg] in the call to water_balance().  it just needs to be tested to make sure this works correctly.

tbohn:
closes #497

### DIFF
--- a/vic/vic_run/src/vic_run.c
+++ b/vic/vic_run/src/vic_run.c
@@ -476,7 +476,8 @@ vic_run(atmos_data_struct   *atmos,
         **********************************************************************/
 
         ErrorFlag = water_balance(lake_var, *lake_con, gp->dt, all_vars,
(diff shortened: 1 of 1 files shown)


jhamman:
Thanks @tbohn. @dgergel is working on getting the travis tests back up and running and we'll restart the travis test manually and make sure everything passes.

PR #500, opened 2016-05-11
Fixed some bugs - related to assuring that image driver has the same behavior as classic driver

yixinmao:
Fixed some bugs
* Some fixes are related to assuring that image driver has the same behavior as classic driver

PR #502, opened 2016-05-12
fixed forgotten loop through the cells for frost_fract

wietsefranssen:
There was no runoff and baseflow in the output. After some investigation I figured out that frost_fract was always zero due to a lack of a loop in `vic_init.c`.
This bugfix just adds this loop through all cells.
Now there is baseflow and runoff again :)


jhamman:
Thanks @wietsefranssen - @yixinmao recently found the same bug and beat you to it on the PR (see #500 for this and a few other fixes).

Issue #381, opened 2016-02-02
ENH: Better multi-file forcing support for the image driver


jhamman:
The image driver currently supports a single multi-file forcing format that appends  the year to the end of the filepath in the format of `forcing/path/prefix.YYYY.nc` where `YYYY.nc` is appended during runtime. For example:

```
#######################################################################
# Forcing Files and Parameters
# netcdf forcing files will be of the form: FORCING1.YYYY.nc
#######################################################################
FORCING1          VIC_test_data/datasets/Stehekin/vic5_setup/Stehekin_image_test.forcings.
```

This requires the user have there forcing files in a very specific configuration. For large model domain as short timesteps, users may find working with the large netCDF files difficult, it would be better if we could come up with a more flexible approach determining which forcing file to use:

Ideally, the user could say something like:

```
FORCING1    VIC_test_data/datasets/Stehekin/vic5_setup/Stehekin_image_test.forcings.YYYY-MM-DD.nc
```

and VIC should be able recognize that the it needs to replace `YYYY` with the year and `MM` with the month and so on. A more pythonic approach may be to use brackets `Stehekin_image_test.forcings.{year}-{month}-{day}.nc`

bartnijssen:
Sure - although there is a large performance hit for using daily files (see issue #380). We'd need to implement this in a way so that we keep the files open while there are still unread records.

jhamman:
Absolutely, we should say just point out in the documentation that the more files you include, the bigger performance hit you're going to have.  I don't see that as a reason not to support smaller files though, it should be up to the user.

PR #505, opened 2016-05-16
Skip correct number of time steps when reading forcing

yixinmao:
- In image driver, updated the code so that when the run start time is later than the forcing start time, it skips the correct number of time steps when reading forcing variables


jhamman:
This looks good.  Thanks @yixinmao.

related: #381

PR #516, opened 2016-06-01
Fixed a unit conversion bug - snow depth, density and SWE

yixinmao:
Fixed a unit conversion bug in vic_run, related to calculating snow depth from snow density and SWE

PR #517, opened 2016-06-02
add restart tests to system testing (and minor refactoring for testing code)

yixinmao:
**NOTE**: this PR incorporates the changes of PR #516; is also related to PR #8 in VIC_sample_data repo. Better to be merged after these two PRs are merged.

Added exact restart tests and minor refactoring for testing code. Specifically:

- [x] Added restart tests including:
      - classic driver and image driver
      - a full permutation of the following options: a) `FULL_ENERGY=TRUE or FALSE`; b) `FROZEN_SOIL=TRUE or FALSE` (when TRUE, use a different soil parameter file where `fs_active=1`); c) for classic driver only, `STATE_FORMAT=ASCII or BINARY`
      - for each of the permutations and drivers, do three VIC runs for the Stehekin test basin: 19490101-19490110 (w/o initial state, this is the full run); 19490101-19490105 (w/o initial state); 19490106-10490110 (w/ initial state from the 19490101-19490105 run). Then check whether the output fluxes for every time step and the states at the end of 19490110 are exactly the same for full run and split runs. Specifically, for classic driver ascii state file format, check whether the fluxes and states are almost the same (due to ascii file restart precision issue); for all the other cases, check whether the outputs are **exactly** the same pixel by pixel
      - Currently, all the restart tests have passed on my machine (yay!)

- [x] Minor refactoring of the testing code:
    - Moved most functions and classes to `test_utils.py`, so that both the main script `run_tests.py` and other scripts including `test_restart.py` and `test_image_driver.py` can use the functionalities in `test_utils.py`

- [ ] Add 'if run on travis' flag in the config file for each individual system test
    - we might not want to run all the tests on travis if it takes too long. A good way to do it (as @jhamman suggested) is to add a flag in the config file for each individual system test, indicating if to be run on travis or not. This can be done in this PR, or in a separate PR later.


jhamman:
@yixinmao - a few style comments and once they've been addressed, I'll give it a full review:

1. There are a number of PEP 8 violations in the python sources in this PR.  Please run these sources through a linter or better yet, use an editor (Atom) that lints while you write.
1. Notably, we want to avoid wild card imports: `from test_restart import *`.
1. The frozen soil parameters for the classic and image drivers belong in the VIC_sample_data repository. Move `Stehekin_soil.FROZEN_SOIL.txt` and `Stehekin_test_params_20160327.FROZEN_SOIL.nc`.

yixinmao:
@jhamman Sounds good! I'll address your suggestions and pin you when I'm done

yixinmao:
@jhamman : I've addressed all your comments. You can go ahead and review it.

tests/run_tests.py
```diff
@@ -1,29 +1,34 @@
 #!/usr/bin/env python
 '''VIC testing command line interface'''
 
-from __future__ import print_function
```
jhamman: this should stay here as it gives us backwards compatibility with python2.

tests/system/system_tests.cfg
```diff
@@ -1,8 +1,214 @@
...
+
+[System-check_classic_nans]
+driver = classic
+test_description = Short simulation to ensure system tests configuration works - classic driver
+global_parameter_file = global.classic.STEHE.txt
+expected_retval = 0
+check = output_file_nans
+
+[System-restart_classic_noFullEnergy_noFrozenSoil]
+# test_description = Exact restart (falseFULL_ENERGY flaseFROZEN_SOIL) - classic driver
```
jhamman: no need to comment out this line.

tests/system/system_tests.cfg
```diff
@@ -1,8 +1,214 @@
...
+# If test_type == exact_restart
+# Full period - start and end date
+# Full period run will start with no initial state file, and the last time step state will be saved
+start_date = 1949,1,1
+end_date = 1949,1,10
+# Dates to split the run
+# E.g., if the full running period is 1949-01-01 to 1949-01-10, and split at 1949-01-05, then the first run would be 1949-01-01 to 1949-01-05, the second run would be 1949-01-06 to 1949-01-10
+# There can be multiple splitting dates
+[[[split_dates]]]
+split_date_1 = 1949,1,5
```
jhamman: I think we can simplify this format a bit.  

```
[System-restart_classic_noFullEnergy_noFrozenSoil]
test_description = Exact restart (falseFULL_ENERGY flaseFROZEN_SOIL) - classic driver
driver = classic
global_parameter_file = global.classic.STEHE.restart.txt
expected_retval = 0
check = exact_restart
[[restart]]
start_date = 1949-01-01
end_date = 1949-01-10
split_dates = 1949-01-04, 1949-01-08
[[options]]
FULL_ENERGY=FALSE
FROZEN_SOIL=FALSE
```

In fact, we don't have to have the `[[restart]]` section at all but since it will only be used when `check = exact_restart`, I think its ok.

tests/run_tests.py
```diff
@@ -316,30 +315,79 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
-                                         test_root=test_dir)
+        # --- if restart test, multiple runs --- #
+        if 'restart' in test_dict:
+            # Set up subdirectories and fill in global parameter options
+            # for restart testing
+            list_global_param =\
+                setup_subdirs_and_fill_in_global_param_restart_test(
+                    s, run_periods, driver, dirs['results'], dirs['state'],
+                    test_data_dir)
+        # --- Else, single run --- #
```
jhamman: simplify the comment style: `# Else, single run`

tests/run_tests.py
```diff
@@ -316,30 +315,79 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
             replacements = OrderedDict()
-        global_param = replace_global_values(global_param, replacements)
+        # --- if STATE_FORMAT is specified, then the specified value (instead
+        # of the one in the global template file) --- #
+        if 'STATE_FORMAT' in replacements:
+            state_format = replacements['STATE_FORMAT']
+        # --- replace global options --- #
+        if 'restart' in test_dict:
+            for j, gp in enumerate(list_global_param):
+                # save a copy of replacements for the next flobal file
```
jhamman: next *global* file

tests/run_tests.py
```diff
@@ -316,30 +315,79 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
+                replacements_cp = replacements.copy()
+                #
```
jhamman: remove or add a comment.

tests/run_tests.py
```diff
@@ -316,30 +315,79 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
             replacements = test_dict['options']
         else:
             replacements = OrderedDict()
-        global_param = replace_global_values(global_param, replacements)
+        # --- if STATE_FORMAT is specified, then the specified value (instead
+        # of the one in the global template file) --- #
+        if 'STATE_FORMAT' in replacements:
+            state_format = replacements['STATE_FORMAT']
+        # --- replace global options --- #
+        if 'restart' in test_dict:
```
jhamman: rather than checking for the `restart` dict, I think it makes more sense to check for the check type `exact_restart`.

tests/run_tests.py
```diff
@@ -351,33 +399,56 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
+                                             logdir=dirs['logs'])
+                    # Check return code
+                    check_returncode(returncode,
+                                     test_dict.pop('expected_retval', 0))
+            else:
+                returncode = vic_exe.run(test_global_file, logdir=dirs['logs'],
+                                         **run_kwargs)
+                # Check return code
+                check_returncode(returncode,
+                                 test_dict.pop('expected_retval', 0))
```
jhamman: pull the `check_returncode` block out of the if/else and for loop.

tests/run_tests.py
```diff
@@ -351,33 +399,56 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
 
-            # Check return code
-            check_returncode(returncode, test_dict.pop('expected_retval', 0))
+            test_complete = True
 
             # check output files (different tests depending on driver)
             if test_dict['check']:
```
jhamman: I think this should be `if 'check' in test_dict:`  I know you didn't write this but it would be worth fixing.

tests/run_tests.py
```diff
@@ -351,33 +399,56 @@ def run_system(config_file, vic_exe, test_data_dir, out_dir, driver):
...
-                fnames = glob.glob(os.path.join(dirs['results'], '*'))
 
                 # Check that the simulation completed for all grid cells
-                if 'complete' in test_dict['check'] and driver == 'classic':
-                    test_classic_driver_all_complete(fnames)
+                if 'complete' in test_dict['check']:
+                    fnames = glob.glob(os.path.join(dirs['results'], '*'))
+                    if driver == 'classic':
+                        test_classic_driver_all_complete(fnames)
```
jhamman: why this change?

tests/system/global.classic.STEHE.restart.FROZEN_SOIL.txt
```diff
@@ -0,0 +1,101 @@
+#######################################################################
+# VIC Model Parameters for Stehekin Basin Sample Classic Driver Setup
+#######################################################################
+
+#######################################################################
+# Simulation Parameters
+#######################################################################
```
jhamman: Comment for all the new global parameter files required for testing. Please remove all comments unless specific to this particular configuration.  I don't like maintaining the comments in these files in 10 places so its easier to just specify the option you are using.  See the examples tests for a good *example*.

tests/system/system_tests.cfg
```diff
@@ -1,8 +1,214 @@
...
+global_parameter_file = global.classic.STEHE.txt
+expected_retval = 0
+check = output_file_nans
+
+[System-restart_classic_noFullEnergy_noFrozenSoil]
+# test_description = Exact restart (falseFULL_ENERGY flaseFROZEN_SOIL) - classic driver
+driver = classic
+global_parameter_file = global.classic.STEHE.restart.txt
+expected_retval = 0
+check = exact_restart
```
jhamman: also check for nans.

tests/test_restart.py
```diff
@@ -0,0 +1,445 @@
...
+    end_date = datetime.datetime(restart_dict['end_date'][0],
+                                 restart_dict['end_date'][1],
+                                 restart_dict['end_date'][2])
+    # --- Identify each of the splitted running period --- #
+    list_split_dates = []
+    for i, split in enumerate(restart_dict['split_dates']):
+        list_split_dates.append(
+            datetime.datetime(restart_dict['split_dates'][split][0],
+                              restart_dict['split_dates'][split][1],
+                              restart_dict['split_dates'][split][2]))
```
jhamman: my comments on how you structure the configuration file will help simplify this section.

tests/test_restart.py
```diff
@@ -0,0 +1,445 @@
...
+import numpy as np
+import datetime
+import os
+import glob
+import xarray as xr
+from test_utils import read_vic_ascii
+
+
+# -------------------------------------------------------------------- #
+def prepare_restart_run_periods(restart_dict, state_basedir, statesec):
```
jhamman: we just removed these comment lines from `run_tests.py` so probably best to do the same here.

tests/test_restart.py
```diff
@@ -0,0 +1,445 @@
...
+        subdirectories under the base directory
+    state_basedir: <str>
+        Base directory of output state results; running periods are
+        subdirectories under the base directory
+    test_data_dir: <str>
+        Base directory of test data
+
+    Returns
+    ----------
+    '''
```
jhamman: returns?

tests/test_restart.py
```diff
@@ -0,0 +1,445 @@
...
+        Base directory of output fluxes results; running periods are
+        subdirectories under the base directory
+    driver: <str>
+        'classic' or 'image'
+    run_periods: <list>
+        A list of running periods. Return from prepare_restart_run_periods()
+
+    Returns
+    ----------
+    None
```
jhamman: no need for a `Returns` section if it doesn't return anything.

tests/test_restart.py
```diff
@@ -0,0 +1,445 @@
...
+                   max().values > 0.0:
+                    raise VICTestError('Restart causes inexact flux outputs '
+                                       '(variable {}) for running period '
+                                       '{} - {} at grid cell {}!'.
+                                       format(var,
+                                              start_date.strftime('%Y%m%d'),
+                                              end_date.strftime('%Y%m%d')))
+                else:
+                    continue
+    return
```
jhamman: return is not needed.

tests/test_restart.py
```diff
@@ -0,0 +1,445 @@
...
+            np.testing.assert_almost_equal(states, states_full_run, decimal=3)
+        # If BINARY state file, check if exactly the same
+        elif state_format == 'BINARY':
+            if states != states_full_run:
+                raise VICTestError('Restart causes inexact state outputs!')
+            else:
+                return
+    elif driver == 'image':
+        for var in ds_states.data_vars:
+            if np.absolute(ds_states[var] - ds_states_full_run[var]).\
```
jhamman: I think you can just use `np.testing.assert_array_equal` here.

tests/test_restart.py
```diff
@@ -0,0 +1,445 @@
...
+               max().values > 0.0:
+                raise VICTestError('Restart causes inexact state outputs'
+                                   '(var {}) !'.format(var))
+            else:
+                continue
+        return
```
jhamman: since you don't share any of the test code here between the classic and image drivers below line 360, just do a single if/elif block.

also, the `else: continue` blocks are not needed in here.

tests/run_tests.py
yixinmao: I put it inside the for loop because I think we need to check returncode for each of the restart runs?

tests/run_tests.py
yixinmao: - For putting the generation of the fnames into the individual checks - I thought this is only needed for some (but not all) tests, so it makes more sense to only do it when needed. But yeah if there are multiple classic driver tests, this step would be done for multiple times with my changes (although I don't think this would take long).
- For putting `if driver == 'classic'` inside - I guess I thought there might be a 'complete' check for image driver at some point?

tests/system/system_tests.cfg
yixinmao: you mean we should check for nans also for every restart test?

tests/system/system_tests.cfg
yixinmao: yeah this is a good suggestion!

yixinmao:
@jhamman : I've addressed most of your comments (and incorporated the travis breaking down issue done in #516), and have replied the rest. Thanks!

PR #529, opened 2016-06-16
added global parameter files for snotel forcings (vic 4.2 and 5.0 comparison)

dgergel:


tests/science/global_param.classic.5.0_snotel.txt
```diff
@@ -0,0 +1,94 @@
...
+
+JULY_TAVG_SUPPLIED    FALSE
+ORGANIC_FRACT         FALSE
+
+VEGLIB                /Users/diana/Dropbox/UW/VIC_test_data/datasets/snotel/parameters/Snotel_VIC.4.1.2_veglib.txt
+VEGPARAM              /Users/diana/Dropbox/UW/VIC_test_data/datasets/snotel/parameters/Snotel_VIC.4.1.2_vegparam.txt
+ROOT_ZONES            3
+
+VEGPARAM_LAI          TRUE
+LAI_SRC               FROM_VEGLIB
```
jhamman: this is different than 4.2

bartnijssen: These hard-coded paths with work for no-one except @dgergel

dgergel: @jhamman, I've changed `LAI_SRC` to `LAI_FROM_VEGPARAM` to be consistent.  

@bartnijssen, indeed - this was an initial debugging commit to ensure that we had the same options set for both versions. I've removed the hard-coded paths.

PR #539, opened 2016-06-27
Setup ameriflux simulations

dgergel:
Added global parameter files for Ameriflux simulations that will be part of the VIC 5 science testing infrastructure. 

- global parameter file for Ameriflux forcings for VIC 4.2

- global parameter file for Ameriflux forcings for VIC 5.0


jhamman:
@dgergel - should this be closed in lieu of #544?

dgergel:
@jhamman, yep, all of this has gone into #544.

PR #558, opened 2016-07-08
check_alloc_status

wietsefranssen:
`check_alloc_status(dvar` should be `check_alloc_status(dvar_filtered`. fixed now


jhamman:
thanks @wietsefranssen

PR #559, opened 2016-07-08
fixed hardcoded mpi path in Makefile

wietsefranssen:
fixed hardcoded mpi path in Makefile


jhamman:
thanks @wietsefranssen

PR #561, opened 2016-07-12
smaller/greater than sign switched

wietsefranssen:


bartnijssen:
Should this just be `!=`? This also seems to be part of PR #564. Since this is a separate bug fix, I'd like to keep it separate from #564

vic/drivers/shared_image/src/check_domain_info.c
```diff
@@ -58,7 +58,7 @@ compare_ncdomain_with_global_domain(char *ncfile)
     // loop over all grid cells and check that the two domains are identical
     for (i = 0; i < global_domain.ncells_total; i++) {
         // mask matches
-        if (ncfile_domain.locations[i].run > global_domain.locations[i].run) {
+        if (ncfile_domain.locations[i].run < global_domain.locations[i].run) {
```
jhamman: I'm not sure what we want to do here.  Both the domain and parameter files have a `mask` variable. This variable currently populates the `locations.run` attribute.  The question is, do we want to be able to turn on and off which grid cells are run?  And if so, which file and variable should this be done in.  

The simplest solution is to say no and assert, as @bartnijssen suggested, that the masks match exactly.

bartnijssen: Oh I see - I was not sure what was being done here either. It would be nice to have a run flag that is separate from the mask, but maybe that is not necessary.

wietsefranssen: I don't see the function of the domain file in case you don't want to be able to turn on an off which grid cells are run.

I think it is usefull to use a file in which you define the cells to run. This "domain-file" should contain only one variable: `run_cells` or someting.

For us it would be even better to be able to use larger param/forcing files than the domain-file. Not only in number of 'run_cells' but also in extend (lat/lon dimension lengths). The advantage is that you dont need to generate new param files and crop-forcing files every time you run on a different part of the total domain. The output of VIC will then be the same as the domain-file (also the lat/lon lengths).

Apart from the discussion: can we just merge this little bug-fix for the time beeing? My test setups do not work without...

bartnijssen:
@wietsefranssen : I'll merge it (or @jhamman can), but you need to resolve the conflicts first

wietsefranssen:
Ah I forgot to update! Done now!

bartnijssen:
@jhamman : Can you merge this after the latest travis problems are resolved?

jhamman:
merged with failing build,  will pass after merge due to #568

PR #562, opened 2016-07-14
j and i index should be switched. Fixed now

wietsefranssen:


jhamman:
nice catch.  will merge once the travis tests pass.

PR #564, opened 2016-07-22
Enhancement/netcdf speed

wietsefranssen:
#520


jhamman:
@wietsefranssen - 

Were you able to see a speedup in the initialization of the image driver with these modifications? I'm going to suggest a different approach, for which @yixinmao may be keen to help with.

1.  Write a new function that opens each input netCDF file (domain, parameters, state, forcings) and sets the `nc_id` attribute in the `nc_file_struct`. This is probably as simple as:

    ```C
        status = nc_open(filenames.parameters, NC_NOWRITE, &(nc->nc_id));
        check_nc_status(status, "Error opening %s", filenames.parameters);
    ```
2.  Write a new function that gets the dimension ids and sizes, and the variable ids for each respective input file (e.g. `get_param_file_nc_info()`).  This will be a custom function for each input file type but will still populate the same `nc_file_struct`.
3. Rework the `get_nc_field_*` functions to use the information in the `nc_file_struct`.  These functions should really just make a single netCDF call, `nc_get_vara_*()`. They should not open/close the file or need to call `nc_inq_varid`.
4.  Write a function that closes all of the input files.  This should go in `vic_finalize()`.

wietsefranssen:
Hi Joe,

this speedup code was mainly been written by @isupit.
For smaller domains (eg Europe/5200 gridcells) the init time improves quite a lot: from 9.46sec to 0.75sec (on average (tested on different machines/nr of cores). if you go even smaller (eg basin level) the speed difference is even larger.
On large domains (eg global) the init time improvement less: 26.84sec to 17.23sec.

Looking to valgrind results I get the idea that opening and closing the netcdf costs the most compared to reading dimensions etc.

I will discuss your suggestions with Iwan

vic/drivers/shared_image/src/check_domain_info.c
```diff
@@ -58,7 +58,7 @@ compare_ncdomain_with_global_domain(char *ncfile)
     // loop over all grid cells and check that the two domains are identical
     for (i = 0; i < global_domain.ncells_total; i++) {
         // mask matches
-        if (ncfile_domain.locations[i].run > global_domain.locations[i].run) {
+        if (ncfile_domain.locations[i].run < global_domain.locations[i].run) {
```
bartnijssen: Should this be `!=`?

bartnijssen:
My guess is that behind the scenes the netcdf library already does what @jhamman  suggests: It maintains a structure with the metadata info for the nc file (identified by ncid) and queries that structure rather than the file. So simple info queries do not go to the file, but are made against this structure. The opening of a file would take a much longer time, because you would have to populate the structure. This is why @wietsefranssen sees a large speedup already. 

So I think that creating an additional structure with all the netcdf metadata may actually duplicate what is already being done in the netcdf library and will not result in much additional speedup and therefore not worth. It would be good to check whether this is true, but for now it would be useful to accept this PR once it's been made up-to-date with the base branch.

It makes sense that the speedup is less for large domains, because the init part does more than parse a netcdf file (allocate memory etc) and the step of parsing metadata info from the netcdf file will be a relatively smaller part of the init time.

bartnijssen:
@wietsefranssen : Can you split out the bug fix that is part of this PR into a separate PR to distinguish it from the feature (netcdf speedup)? I am talking about commit a2d89a , although maybe that has already been accepted separately.

wietsefranssen:
just merged to latest 'develop' and seperated bugfix #561

jhamman:
@yixinmao - the task we discussed yesterday should probably superseded this PR. Just want to make sure you were aware of this existing work.

yixinmao:
@jhamman Thanks Joe - I did actually missed this PR. Do we want to just open a new PR later? Or do we want to actually continue on from this one?

jhamman:
@yixinmao - Go ahead and just open a new PR.

yixinmao:
@jhamman Sounds good. And I briefly looked at this PR and I think the way I started doing is quite similar to what they did.

bartnijssen:
Superseded by #684. Since that one is based on a more recent version of the code I am going with that one.

PR #567, opened 2016-07-26
Add MPI test

yixinmao:
Add MPI multiprocessing test
- test whether the flux and state results using different number of processors are exactly identical for image driver
- add instruction on how to run VIC with multiple processors


bartnijssen:
@yixinmao: Can you provide a bit more detail on the error you are getting?

yixinmao:
@bartnijssen This failing is related to the MPI part in VIC run, and I haven't figured out what's going on and @jhamman is going to take a look as well. Specifically, on travis, when we do not write state output files, VIC runs successfully; but when we do output state files and with multiple processors, VIC gives an error. But everything runs successfully locally on hydra. So it still needs investigation.

yixinmao:
@jhamman I've merged your bug fix and cleaned up. Ready to merge.

PR #569, opened 2016-07-29
Fix bug in air density calculation

yixinmao:
Fixed bug in air density calculation


jhamman:
@yixin - can you point us to the line in VIC 4.2 that had the correct formulation for air density.  Also, do you have a before and after plot comparing 4.2 and 5?

yixinmao:
@jhamman  the line in VIC4.2 that calculates air density is here: https://github.com/UW-Hydro/VIC/blob/support/VIC.4.2.d/src/initialize_atmos.c#L1156
I have comparison plots for each grid cell in the Stehekin 10-day example for all output variables. For air density, they look identical after the fix.

jhamman:
looks good.  Thanks @yixinmao!

PR #570, opened 2016-08-04
Fixed bugs related to ice heat capacity

yixinmao:
- In vic_run code, fixed bugs that confused volumetric heat capacity
      and specific heat capacity related to ice
- We can decide if the value of `CONST_CPICE` (specific heat capacity of ice) is reasonable; but it should not make a huge difference on results

### DIFF
--- a/vic/vic_run/include/vic_physical_constants.h
+++ b/vic/vic_run/include/vic_physical_constants.h
@@ -108,6 +108,8 @@
 #define CONST_CPFW 4.188e3  /**< specific heat of fresh h2o ~ J/kg/K */
 #define CONST_CPFWICE 4.2e3  /**< specific heat of fresh h2o ~ J/kg/K */
 #define CONST_CPICE 2.11727e3  /**< specific heat of fresh ice ~ J/kg/K */
+/**< volumetric heats */
+#define CONST_VCPICE_WQ (CONST_CPICE * CONST_RHOFW)  /**< heat capacity of fresh ice per volume of water equivalent ~ J/m^3/K */
 /**< latent heats */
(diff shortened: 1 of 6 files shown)


jhamman:
@yixinmao - thanks for rooting this out.  This is huge. I'm almost certainly to blame for introducing this bug so thanks for cleaning up my mess.

I wonder if we should define a variable in `vic_physical_constants.h` that holds the value of `CONST_CPICE * CONST_RHOFW`.  It seems you use that over and over again in this fix.

bartnijssen:
Yes - if they are physical constants we should handle them that way (include in `vic_physical_constants.h`. 

@yixinmao : Can you make that change and make sure the branch is up-to-date with the base? Then we can merge.

yixinmao:
@jhamman @dgergel - Done!

vic/vic_run/include/vic_physical_constants.h
```diff
@@ -108,6 +108,8 @@
 #define CONST_CPFW 4.188e3  /**< specific heat of fresh h2o ~ J/kg/K */
 #define CONST_CPFWICE 4.2e3  /**< specific heat of fresh h2o ~ J/kg/K */
 #define CONST_CPICE 2.11727e3  /**< specific heat of fresh ice ~ J/kg/K */
+/**< volumetric heats */
+#define CONST_VCPICE_WQ (CONST_CPICE * CONST_RHOFW)  /**< heat capacity of fresh ice per volume of water equivalent ~ J/m^3/K */
```
bartnijssen: What's the `WQ` at the end stand for?

bartnijssen: OK - water equivalent - I'll just answer my own question. I'll go ahead and merge.

PR #573, opened 2016-08-06
Fixed a bug related to VP - to be consistent with VIC4.2

yixinmao:
- When calculated VPD < 0, now resets VPD to zero, and VP to calculated saturated vapor pressure (this is consistent with VIC4.2)

### DIFF
--- a/vic/drivers/classic/src/vic_force.c
+++ b/vic/drivers/classic/src/vic_force.c
@@ -138,6 +138,10 @@ vic_force(force_data_struct *force,
             force[rec].vp[i] = forcing_data[VP][uidx] * PA_PER_KPA;
             // vapor pressure deficit in Pa
             force[rec].vpd[i] = svp(force[rec].air_temp[i]) - force[rec].vp[i];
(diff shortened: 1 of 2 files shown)


jhamman:
This looks good to me.  @bartnijssen and @tbohn, any issue with merging this?

tbohn:
Looks good to me.  Vastly preferable to having negative VPD!

I suppose this would only ever occur if we had observations of VP or other metrics of humidity, particularly if we didn't also have simultaneous observations of air temperature (e.g., spline-interpolated hourly air temperature that isn't consistent with observed VP).  If we're estimating VP via MTCLIM or getting VP from an atmospheric model, VP would be consistent with air temperature, right?  So I'd think that this would be a relatively uncommon situation...

yixinmao:
@tbohn Thanks Ted for reviewing! It turned out that even if we derived both air T and humidity via MTCLIM, because of the limited precision of forcing data (at least in classic driver), negative VPD sometimes still occurs before this PR. The slightly negative VPD somehow results in visible change in model results...

PR #574, opened 2016-08-09
Fixed bugs related to volumetric heat capacity of water

yixinmao:
- Fixed bugs where volumetric heat capacity of water (instead of specific heat) should be used
- **NOTE**: it seems that there is a bug in `VIC4.2` - the hardcoded number `4186.8` here should be volumetric heat of water (which is 4186.8e3):  https://github.com/UW-Hydro/VIC/blob/support/VIC.4.2.d/src/func_canopy_energy_bal.c#L262 . Other places in `VIC4.2` related to volumetric heat of water appear to be correct)

PR #575, opened 2016-08-10
Fixed a bug - volumetric heat capacity of water 

yixinmao:
- Volumetric heat capacity of water  should be used in `func_canopy_energy_bal.c` instead of specific heat


jhamman:
see #574 for further discussion on this bug fix.

PR #577, opened 2016-08-13
updated paths to agree with directory structure of the VIC5_test_data…

dgergel:
This PR updates science test paths so that they agree with the VIC5 science test data FTP server directory structure.


jhamman:
thanks @dgergel

PR #588, opened 2016-08-18
Point DOI to latest version on zenodo

bartnijssen:


jhamman:
ah, this is just great.  I was wondering how to do this.  Thanks.

bartnijssen:
I think they changed how they do this on zenodo, cause this was now the default link that they provided (i.e. no work on my end).

PR #595, opened 2016-08-20
Updated vic_force in image driver - now input forcing units are consistent with classic driver

yixinmao:
Previously, the two drivers require different sets of units for input forcings, which is confusing. Now, the required forcing units for all the 7 required forcing input variables are consistent.


jhamman:
This looks good to me.  Thanks.

Issue #376, opened 2016-01-27
ENH: Build VIC drivers using Autotools/Automake


jhamman:
It would be nice to have a single command-line utility to handle building any of the VIC drivers.  This could be done using GNU Autotools/Automake. This might look something like this:

```Bash
./configure --driver=image \
            --disable-openmp \
            --with-netcdf=yes \
            --prefix=./ \
            --enable-extension=rvic
```
As we add dependencies and build out the extensions in VIC 5, we can put most of the build time decisions in the configure scripts, allowing the VIC source code to stay mostly clear of build time directives.

bartnijssen:
:+1:

PR #601, opened 2016-08-25
Add CMake file.

Sibada:
It may be easier to compile VIC,set the path of dependences and choose which driver to build. And now only support Classic and Image drivers.
```bash
cd vic
mkdir build
cd build
cmake -D CLASSIC=OFF -D IMAGE=ON -D SHARED_LIB=ON -D INSTALL=ON ..
make -j8
sudo make install
...
```


jhamman:
Thanks @Sibada.

This looks like an interesting addition and is certainly related to #376. CMAKE may be the way to go but we should discuss with the larger group (@UW-Hydro/vic-developers, @tbohn, @wietsefranssen).

jhamman:
closed #605

wietsefranssen:
I just red the CMake discussions. Looks good (although I am not good in cmake yet)! 
I also think it is important to make it more generalised and avoid for example hardcoded depenency paths.
For us it would be very usefull to also add options like `--enable-extension=rvic`(see #376) in VIC5.1.

PR #604, opened 2016-08-25
Fixed a broken link in image driver documentation

hhelgason:
Fixed a link to the parameter conversion notebook in the image driver documentation.

### DIFF
--- a/docs/Documentation/Drivers/Image/Params.md
+++ b/docs/Documentation/Drivers/Image/Params.md
@@ -1,6 +1,6 @@
 # VIC Model Parameters
 
-The Image Driver uses the [NetCDF](http://www.unidata.ucar.edu/software/netcdf/) file format for its input model parameters. It is possible to convert the VIC ASCII style parameters to this format. We have put together an example ([Tutorial](Ascii_to_NetCDF_params.md) and [Ipython Notebook](../../../../samples/notebooks/example_reformat_vic4_parameters_to_vic5image.ipynb)) that provide examples of how to do this conversion. Our example uses the `tonic` [Python](https://www.python.org/) Package.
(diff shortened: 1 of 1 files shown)


docs/Documentation/Drivers/Image/Params.md
```diff
@@ -1,6 +1,6 @@
 # VIC Model Parameters
 
-The Image Driver uses the [NetCDF](http://www.unidata.ucar.edu/software/netcdf/) file format for its input model parameters. It is possible to convert the VIC ASCII style parameters to this format. We have put together an example ([Tutorial](Ascii_to_NetCDF_params.md) and [Ipython Notebook](../../../../samples/notebooks/example_reformat_vic4_parameters_to_vic5image.ipynb)) that provide examples of how to do this conversion. Our example uses the `tonic` [Python](https://www.python.org/) Package.
+The Image Driver uses the [NetCDF](http://www.unidata.ucar.edu/software/netcdf/) file format for its input model parameters. It is possible to convert the VIC ASCII style parameters to this format. We have put together an example ([Tutorial](Ascii_to_NetCDF_params.md) and [Ipython Notebook](https://github.com/UW-Hydro/VIC/blob/develop/samples/notebooks/example_reformat_vic4_parameters_to_vic5image.ipynb)) that provide examples of how to do this conversion. Our example uses the `tonic` [Python](https://www.python.org/) Package.
```
bartnijssen: @jhamman : Is there a way to do this so that it is not an absolute link?

jhamman: Not really.  We could convert the notebook into markdown, and include that as page.  I would prefer to let github render the notebook for us and live with the fact that there may be a discontinuity in the version of the docs and the notebook.

PR #605, opened 2016-08-26
ENH: Use CMake to build VIC drivers

Sibada:
I set a series of probably path of MPI and netCDF for CMake to search them. That's also the way MXNet take.


jhamman:
@Sibada, this PR looks quite interesting.

I'm setting this PR as an enhancement that would go into VIC 5.1.  I'm not really interested in managing two build systems (e.g. GMake and CMake) so we need to make sure we achieve at least the current build functionality with CMake that we currently have with GMake. It would probably be good for you to review the makefiles ([Image](https://github.com/UW-Hydro/VIC/blob/develop/vic/drivers/image/Makefile), [Classic](https://github.com/UW-Hydro/VIC/blob/develop/vic/drivers/classic/Makefile), [CESM](https://github.com/UW-Hydro/VIC/blob/develop/vic/drivers/cesm/Makefile)) for the existing drivers to get an idea of what functionality we currently have. Before we merge this, the following minimum improvements would need to be made.

#### Pass the following to the c compiler
- [ ] GIT_VERSION (e.g. `$(shell git describe --abbrev=4 --dirty --always --tags)`)
- [ ] UNAME_S (e.g. `$(shell uname -s)`)
- [ ] HOSTNAME (e.g. `$(shell uname -n)`)

#### Configuration and build options
- [ ] C compiler (default to mpicc)
- [ ] netCDF flags (NC_FLAGS default to `$(shell nc-config --cflags)`, NC_LIBS default to `$(shell nc-config --libs)`)
- [ ] LOG_LVL (default to 5, set on command line)
- [ ] Optimization flags
- [ ] Install directory (default to `VIC/bin)`

#### Testing
- [ ] Integrate with [TravisCI](https://github.com/UW-Hydro/VIC/blob/develop/.travis.yml)
- [ ] Integrate with [Docker](https://github.com/UW-Hydro/VIC/blob/develop/ci/Dockerfile)

#### Misc.
- [ ] Remove all hard-coded paths (e.g. `/usr/local/include`)
- [ ] Document the new method of building VIC (e.g. [here](https://github.com/UW-Hydro/VIC/blob/master/docs/Documentation/Drivers/Image/RunVIC.md))



cc @UW-Hydro/vic-developers, @tbohn, @wietsefranssen


xref: #376, #601

Sibada:
I found that the characteristics of travis ci is not the same as in my imagination.

vic/CMakeLists.txt
```diff
@@ -0,0 +1,207 @@
...
+            /usr/include/mpich
+            /usr/local/include/mpich
+            /usr/local/mpich/include
+       )
+    set(OPENMPI_INCLUDE_SEARCH_PATHS
+            $ENV{HOME}/openmpi/include
+            /usr/include/openmpi
+            /usr/local/include/openmpi
+            /usr/local/openmpi/include
+       )
```
jhamman: I don't think we need any of these hard coded paths. The `mpicc` compiler should include all include and linking flags necessary to compile VIC.

vic/CMakeLists.txt
```diff
@@ -0,0 +1,207 @@
...
+            /usr/lib
+            /usr/lib/netcdf
+            /usr/lib/x86_64-linux-gnu
+            /usr/lib64
+            /usr/local/lib
+            /usr/local/lib64
+            /usr/local/netcdf/lib)
+
+    find_path(NETCDF_INCLUDE_DIR NAMES netcdf.h PATHS ${NETCDF_INCLUDE_SEARCH_PATHS})
+    find_path(NETCDF_LIB_DIR NAMES libnetcdf.so PATHS ${NETCDF_LIB_SEARCH_PATHS})
```
jhamman: Remove hard coded paths.  Take a look at the existing image driver makefile for an example on how to do this.  `nc-config -- libs` and `nc-config --cflags` should give you the necessary information.  Users should be able to overwrite those though.

vic/CMakeLists.txt
```diff
@@ -0,0 +1,207 @@
...
+# ============================================================================
+
+# Cmake configurations options. ==============================================
+option(CLASSIC      "Compile VIC Classic Driver"    OFF)
+option(IMAGE        "Compile VIC Image Driver"      ON)
+option(SHARED_LIB   "With Vic shared library"       OFF)
+option(INSTALL      "Install VIC to system path"    OFF)
+# ============================================================================
+
+set(CMAKE_INSTALL_PREFIX /usr/local/vic) # Path to install VIC shared library.
```
jhamman: most users will not have write permissions here.  I would suggest defaulting to either the driver directory (e.g. `VIC/vic/drivers/image/`) or `VIC/bin/`.

ci/cmake.travis
```diff
@@ -0,0 +1,100 @@
...
+
+function vic_after_success {
+    echo vic_after_success
+    echo "Success!"
+}
+
+function vic_after_failure {
+    echo vic_after_failure
+    echo "Test failed -- please review the log"
+}
```
jhamman: This is a good start on the travis integration.

However, I'd prefer not supporting both GNU-Make and CMake at the same time. So rather than just adding a new travis build as you have, we should swap out all existing builds for CMake.

jhamman:
@Sibada - do you have plans to come back to this or should we close this PR?

bartnijssen:
No response - will close PR for now

PR #623, opened 2016-09-01
Update state file timestamp

yixinmao:
- *VIC code change* - shifted output state file timestamp to the end of a certain timestep (this timestamp is still consistent with what user specifies in the global file)
- *Test code change* - made corresponding changes in the system test code, as well as all global files in the test directory
- "Doc update" - update docs related to output state file


yixinmao:
This PR should replace and close PR #618

jhamman:
This looks great.  I have a few style updates that I'll issue a separate PR for. Thanks @yixinmao for the quick turn around on this.

PR #629, opened 2016-09-01
Fixed NYEARS frequency when there are leap years

yixinmao:
Previously NYEARS frequency resulted in `day_in_year = 366` when advance from a leap year to a non-leap year. Fixed this problem by resetting `day_in_year` after advancing year(s).

### DIFF
--- a/vic/drivers/shared_all/src/vic_time.c
+++ b/vic/drivers/shared_all/src/vic_time.c
@@ -859,13 +859,13 @@ time_delta(dmy_struct        *dmy_current,
             // if dmy_next.year is not leap year but date is Feb 29 !)
             make_lastday(global_param.calendar, dmy_next.year, lastday);
             dmy_next.day_in_year = 0;
-            for ( i = 1; i <= MONTHS_PER_YEAR; i++ ) {
-                if ( i < dmy_next.month ) {
-                    dmy_next.day_in_year += lastday[i-1];
-                }
-                else if ( i == dmy_next.month ) {
+            for ( i = 0; i < MONTHS_PER_YEAR; i++ ) {
+                if ( (i+1) == dmy_next.month ) {
                     dmy_next.day_in_year += dmy_next.day;
                     break;
+                }
+                else {
+                    dmy_next.day_in_year += lastday[i];
                 }
             }
         }
--- a/vic/drivers/shared_all/src/vic_time.c
+++ b/vic/drivers/shared_all/src/vic_time.c
@@ -820,6 +820,8 @@ time_delta(dmy_struct        *dmy_current,
 
     double                     td, a, b;
     dmy_struct                 dmy_next;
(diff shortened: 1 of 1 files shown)


vic/drivers/shared_all/src/vic_time.c
```diff
@@ -851,7 +853,21 @@ time_delta(dmy_struct        *dmy_current,
             }
         }
         else if (freq == FREQ_NYEARS) {
+            // advance year
             dmy_next.year += n;
+            // re-calculate day_in_year (NOTE: currently will have problem
+            // if dmy_next.year is not leap year but date is Feb 29 !)
```
bartnijssen: Clarify comment. It's unclear what `currently` refers to. Is that before or after the fix. If before, then it does not belong here (since you can put this info in the commit message). If after then it seems we still have a problem.

vic/drivers/shared_all/src/vic_time.c
```diff
@@ -851,7 +853,21 @@ time_delta(dmy_struct        *dmy_current,
...
+            make_lastday(global_param.calendar, dmy_next.year, lastday);
+            dmy_next.day_in_year = 0;
+            for ( i = 1; i <= MONTHS_PER_YEAR; i++ ) {
```
bartnijssen: C-counters typically run from `0` through `< LIMIT`. That also gets around the `i-1`

vic/drivers/shared_all/src/vic_time.c
```diff
@@ -851,7 +853,21 @@ time_delta(dmy_struct        *dmy_current,
...
+                if ( i < dmy_next.month ) {
+                    dmy_next.day_in_year += lastday[i-1];
+                }
+                else if ( i == dmy_next.month ) {
+                    dmy_next.day_in_year += dmy_next.day;
+                    break;
+                }
+            }
```
bartnijssen:
```C
for (i = 0; i <= MONTHS_PER_YEAR; i++) {
    if ((i+1) == dmy_next.month) {
        dmy_next.day_in_year += dmy_next.day;
        break;
     } 
    else {
        dmy_next.day_in_year += lastday[i];
    }
}
```

vic/drivers/shared_all/src/vic_time.c
yixinmao: @bartnijssen It is after the fix. This note is for the purpose of reminder for future code improvement. There will only be a problem if a simulation starts on Feb 29 of a leap year, and yearly aggregation/output is used for AGGFREQ or HISTFREQ (which should be rare case of usage...). We can definitely fix this potential problem also, but it requires some more logics. For now I just added an error message below saying that this is a simulation time case that VIC does not support now, if that's fine.

bartnijssen: Please open an issue for that I'm OK with the error message for now

PR #635, opened 2016-09-09
Fixed image driver history file name

yixinmao:
Fixed image driver history file name timetamp - now is timstep-beginning


jhamman:
@yixinmao - This looks good. Can you add a note in the `ReleaseNotes`?

yixinmao:
@jhamman - How should I add the change to `ReleaseNotes` for this case, since I guess it's not supposed to be under `VIC 5.0.0`, but maybe also not under `Known Issues in Current Release` ?

yixinmao:
Release notes added, thanks @jhamman !

PR #639, opened 2016-10-05
Fixed a bug when calculating steps of forcing to skip 

yixinmao:
When calculating number of steps to skip in the forcing data, should round to integer first before converting to int type to prevent possible 1-timestep wrong offset (due to precision)


yixinmao:
@jhamman done, thanks!

PR #642, opened 2016-10-08
Feature/debug cesm driver

dgergel:


vic/drivers/shared_image/src/vic_store.c
```diff
@@ -25,12 +25,14 @@
...
 
 #include <vic_driver_shared_image.h>
+#include <vic_driver_cesm.h>
 
 /******************************************************************************
  * @brief    Save model state.
  *****************************************************************************/
 void
 vic_store(dmy_struct *dmy_current,
+          case_metadata *cmeta,
```
jhamman: we don't want to pass `case_metadata` in to `vic_store`.  Instead, you should set `filenames.statefile = cmeta->caseid`.

vic/drivers/shared_image/src/vic_store.c
```diff
@@ -25,12 +25,14 @@
  *****************************************************************************/
 
 #include <vic_driver_shared_image.h>
+#include <vic_driver_cesm.h>
```
jhamman: this file is also compiled with the image driver so you need to find a solution that does not require the cesm driver.  I think that all you need is the caseid which can be stored as `filenames.statefile`.

dgergel: Yep - I'm doing this now in the `vic_cesm_run` function. Take a look at the way I'm reassigning `filenames.statefile` though - it didn't work the way you suggested (I don't know why, since I'm new to this) and this was a way that worked.

vic/drivers/cesm/src/cesm_interface_f.F90
```diff
@@ -77,11 +77,12 @@ END FUNCTION vic_cesm_init
   !> @brief   Run Interface
   !--------------------------------------------------------------------------
   INTERFACE
-     INTEGER(C_INT) FUNCTION vic_cesm_run(vclock) BIND(C, name='vic_cesm_run')
+     INTEGER(C_INT) FUNCTION vic_cesm_run(vclock, cmeta) BIND(C, name='vic_cesm_run')
        USE, INTRINSIC :: ISO_C_BINDING
        USE vic_cesm_def_mod
        IMPLICIT NONE
        TYPE(vic_clock), INTENT(in) :: vclock
+       TYPE(case_metadata), INTENT(in) :: cmeta
```
jhamman: you can roll back these changes as well.

vic/drivers/shared_image/src/vic_store.c
```diff
@@ -48,25 +48,44 @@ vic_store(dmy_struct *dmy_current,
...
+    time_num = date2num(global_param.time_origin_num, dmy_current, 0, 
+                        global_param.calendar, global_param.time_units);
+    end_time_num = time_num + offset; 
+
+    // allocate dmy struct for end of current time step 
+    num2date(global_param.time_origin_num, end_time_num, 0., 
+                             global_param.calendar, global_param.time_units, 
+                             &end_time_date);
+    
+
```
jhamman: This block looks good but can you move it into the `if (mpi_rank == VIC_MPI_ROOT)` block below.

vic/drivers/cesm/src/cesm_interface_c.c
```diff
@@ -96,9 +96,16 @@ vic_cesm_init(vic_clock     *vclock,
...
+    // initialize forcings
+    vic_force();
+
     // initialize output structures
     vic_init_output(&dmy_current);
 
+    // assign case name to state file name
+    strncpy(filenames.statefile, trim(cmeta->caseid),
+            sizeof(filenames.statefile));
+
```
jhamman: please put this in `vic_cesm_start()`.  As you can see, we don't really do anything in this top level function except call other functions and I'd like to keep it that way.

vic/drivers/shared_image/src/vic_store.c
dgergel: Actually, on second thought, the timestamp will be incorrect if it's set at initialization and not at the time at which we write the state file. But I still think it seems like this shouldn't be in `vic_store`, since that's a shared_image function.

bartnijssen:
@jhamman : Can you review and merge when ready?

vic/drivers/cesm/src/cesm_interface_c.c
jhamman: this still needs to be addressed.

vic/drivers/shared_image/src/vic_store.c
jhamman: This needs to be addressed still.

bartnijssen:
@dgergel : Please complete so @jhamman can merge

dgergel:
@jhamman I'll tie this up today.

vic/drivers/cesm/src/cesm_interface_c.c
dgergel: Updated.

vic/drivers/shared_image/src/vic_store.c
dgergel: Updated.

PR #645, opened 2016-10-14
Added check to ensure mask variable in the input domain file is integer type

yixinmao:
Added check to ensure mask variable in the input domain file is integer type; otherwise raise an error.

### DIFF
--- a/vic/drivers/shared_image/src/get_nc_var_type.c
+++ b/vic/drivers/shared_image/src/get_nc_var_type.c
@@ -1,7 +1,7 @@
 /******************************************************************************
  * @section DESCRIPTION
  *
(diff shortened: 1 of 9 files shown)


vic/drivers/shared_image/src/get_nc_var_type.c
```diff
@@ -0,0 +1,60 @@
+/******************************************************************************
+ * @section DESCRIPTION
+ *
+ * Get netCDF variable attribute.
```
jhamman: ~~attribute~~ --> type

yixinmao:
@jhamman all comments addressed and domain file docs added.

Issue #649, opened 2016-11-05
Seems `run_cell` not works?


Sibada:
I had set `run_cell` to 1 at basins in different scale and other gridcells set to 0 but they cost nearly a same time, and each cell output has data. Or it should be set in domain file?

bartnijssen:
Thanks for filing the issue. Please report which version of VIC you are running, what the driver is you are using, etc. As is, it's difficult to do anything with the message.

yixinmao:
@Sibada If you are running VIC5.0.0 image driver, currently the mask variable in the input domain file controls running cells. `run_cell` variable in the image parameter file is not used. Thanks for the issue and we will update the documentation (but as @bartnijssen said, reporting sufficient information inlcuding the version and driver you are using, etc. will help us better handle issues and questions in the future).

bartnijssen:
@yixinmao : Can this be closed or should this remain open?

yixinmao:
@bartnijssen We've decided on a solution, but I haven't implemented changes yet (I looked at it and it needs some work - will do so soon). So maybe keep it open?

bartnijssen:
@yixinmao : Just pinging you to keep this alive and on your radar

yixinmao:
@bartnijssen Thanks Bart! Sorry I've been pushing this back... will work on it soon!

Sibada:
0_0

Sibada:
In fact, I solve this at the next day but I forgot to response here...
I'm using the VIC5 for my research and the paper has nearly been finished, are you have some new papers special for the version 5 could be referenced?

jhamman:
@Sibada - 

We If you have a fix you'd like to share, please open a pull request. @yixinmao can help push that through.

We do have a paper describing the VIC 5 development and model release.  It is also nearing submission.  The citation will be:

> Hamman, J., B. Nijssen, T. Bohn, Y. Mao, and D. Gergel: The Variable Infiltration Capacity (VIC) Model, Version 5.0 - improvements and new applications. In preparation, to be submitted to Geoscientific Model Development - 2017.

yixinmao:
@bartnijssen : I discussed with @jhamman and @dgergel

yixinmao:
[...]

PR #650, opened 2016-11-08
Update image driver docs - `run_cell` not used in param file, removed from the docs

yixinmao:
This PR will close issue #649


bartnijssen:
I don't understand the change - wouldn't the proper fix be to use `run_cell` to determine whether a cell is run or not?

yixinmao:
@bartnijssen : right now there are two places that could specify whether a cell is run or not - `mask` variable in the input domain file, and `run_cell` variable in the parameter file. Currently in the code, we are only using `mask` to determine if a cell is run or not, and `run_cell` is not used at all. And I think it makes sense to run the cells based on the domain file, that's why I made the change, since the current documentation is misleading.

But another way to do it is we completely use `run_cell` to control running a cell or not. I guess the benefit of doing so is that if the user has a large domain (thus the 'mask' variable), but s/he wants to turn some grid cells on and off, s/he can adjust `run_cell` in the parameter file to do so rather than changing the `mask` variable in the domain file. If this way is preferred, we can make code changes to achieve so.

bartnijssen:
To me these two things signify something different. 

`mask` in the domain file simply designates all the cells that are part of a model domain. For example, for RASM this domain never changes. 
`run_cell` makes it clear which cells in the domain should actually be simulated.

This is not the same thing. We are in the habit of having masks that match exactly with the information that we have in other fields (so all fields outside the mask are `Fill_Value`), but that just happens to be the way in which we implement most of our simulations. In that case a separate `mask` is not necessary. But I can imagine cases in which you have values over the entire domain (in the input fields), in which case you cannot easily reconstruct the `mask` if you just want to run over a small subset, which is why we have `run_cell`.

So the solution proposed here is not the one I want us to use. We should use `run_cell` in the code to determine whether a cell needs to be run, not `mask`.

bartnijssen:
Given the above I am closing the PR. Fixing it the other way (and use `run_cell`) should be an easy fix as well.

bartnijssen:
It's fine with me if we add `run_cell` to the domain file as part of this (not sure where it currently lives).

yixinmao:
@bartnijssen OK that makes sense. I do agree that the second way is more reasonable and flexible for users. I'll make the changes later.

PR #651, opened 2016-11-10
fix deprecated options MIN_RAIN_TEMP and MAX_SNOW_TEMP

dgergel:


yixinmao:
@dgergel looks good to me.

PR #652, opened 2016-11-11
fixed bug in writing statefile for 2-d lat/lon dimensions

dgergel:
2-D lats/lons were being written to state file as fill values rather than actual lats/lons, this PR fixes that.


dgergel:
cc @yixinmao, @jhamman for review.

yixinmao:
@dgergel looks good to me.

PR #657, opened 2016-12-08
Fix dz_node and node_depth in image statefile

yixinmao:
In image driver, `dz_node` and `node_depth` was assumed to be not spatially-distributed, which was wrong. Now output spatially distributed `dz_node` and `node_depth` in the output state file; also, when reading in an initial state file, check these variables at all grid cells to see if they match the values calculated from soil parameters.

### DIFF
--- a/docs/Development/ReleaseNotes.md
+++ b/docs/Development/ReleaseNotes.md
@@ -38,6 +38,10 @@ To check which release of VIC you are running:
 
 	Before the fix, the input arguments to function `make_lastday` are sometimes in a wrong order. The bug caused error when trying to write state file on a leap day.
  
+6. Fixed a bug related to `dz_node` and `node_depth` variables in image driver output state file ([GH#657](https://github.com/UW-Hydro/VIC/pull/657))
+
+	Before the fix, `dz_node` and `node_depth` in image driver output state file were not spatially distributed, which was wrong. Now these two variables are spatially distributed in the output state file.
+ 
 ------------------------------
 
 ## VIC 5.0.0 [![DOI](https://zenodo.org/badge/7766/UW-Hydro/VIC.svg)](https://zenodo.org/badge/latestdoi/7766/UW-Hydro/VIC)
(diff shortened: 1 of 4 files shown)


bartnijssen:
Do we need to store them, or do we just make sure they are calculated every time the model starts?

yixinmao:
@bartnijssen Currently they are already calculated every time the model starts, and then in image driver it further checks if the calculated values match those in the initial state file (classic driver does not do the check; it always use those saved in the initial state file). I guess for model running purpose, we don't have to store them; for practical purpose (user wants to know node depth info), it might be informative - or we could write it out to history file if user outputs node-related variables.

jhamman:
@bartnijssen and @yixinmao - here's my perspective, we want to write/read coordinate variables like `dz_mode` for two reasons:

1. It is useful to be able to inspect a statefile as a standalone object. Since `dz_node` is not a model parameter or a variable that will exist in the domain file, user would have to reproduce the behavior in the VIC if they wanted to compare a state variable to its VIC calculated coordinate.
2. Many state variables are not transferable when model options are changed. VIC can calculate `dz_node` in two ways (linear or exponential), both using the same number of nodes. If we were to stop writing out the node depths, and a user were to change the way the nodes are distributed via a global parameter option, the incorrect state would be applied to their model. In fact, this is the primary use case right now.  VIC calculates its node depths at startup and then compares them to those in the statefile, exiting if they differ.

yixinmao:
@jhamman So after the fix in this PR, image driver is doing what you proposed. Should we add the same check for classic driver as well?

jhamman:
@yixinmao - no.  The benefit of using netCDF to store the image driver state is there is a clear and concise way to keep metadata with the state variables.  The classic driver doesn't really fit that bill so I would not suggest extending this beyond the image driver.

yixinmao:
@jhamman OK - the same two variables (`dz_node` and `node_depth`) are already (and have always been) stored in state file in classic driver; and they are recalculated at the beginning of VIC run as well - but just not checked (values in the state file are directly used).

bartnijssen:
@yixinmao : Is this one ready to merge or are we waiting for something (other than me)

yixinmao:
@bartnijssen this should be ready to merge once someone takes a quick review. @jhamman @dgergel could you take a brief look? Most changes shown are due to indenting.

jhamman:
LGTM. After @dgergel takes a look, I think this can go in.

dgergel:
@yixinmao and @jhamman this looks good to me. @bartnijssen, ready to merge.

PR #662, opened 2017-01-05
Fix run_cell bug in image driver

yixinmao:
Before this PR, active cell is controlled by `mask` variable in the domain file in image driver. Now after the fix in this PR, `run_cell` variable in the parameter file controls active cells. Specifically, we
re-organized `get_global_domain` function to:
    1) use `run_cell`variable from the parameter file to determine
active cell (instead of the `mask` variable in the domain file);
    2) lat and lon coordinates loading is put into a separated function;
    3) the step of checking whether the latlons in the parameter file
match the domain file is moved to early on in the `get_global_domain`
function.

Some docs updates related to this change is also incorporated. The Stehekin sample dataset is updated to be unambiguous in the `VIC_sample_data` repo (https://github.com/UW-Hydro/VIC_sample_data/pull/16).


yixinmao:
@jhamman this PR is ready for review.

vic/drivers/shared_image/src/get_global_domain.c
```diff
@@ -103,17 +118,77 @@ get_global_domain(char          *nc_name,
...
+
+
+/******************************************************************************
+ * @brief    Get lat and lon coordinates information from a netCDF file and
+             store in nc_domain structure
+ *****************************************************************************/
+void
+get_nc_latlon(char          *nc_name,
+              domain_struct *global_domain,
+              domain_struct *nc_domain)
```
jhamman: can you help me understand how these two domain structs differ and what you intend for `nc_domain` to be. Perhaps some more documentation in the header is appropriate.

jhamman: I'd be interested to hear what @bartnijssen thinks is best here.

bartnijssen: I find this confusing as well. It makes the function do more than it needs to anyway. If you are sure that the domain information from `global_domain` can simply be copied to `nc_domain`, then do so explicitly before you enter the function. For example, write a small `copy_domain(domain_struct *domain_from, domain_struct *domain_to)` function. Call that first and then call `get_nc_latlon(char *nc_name, domain_struct *nc_domain)`. That way there is no hidden beahvior (the copy that happens as a side effect in `get_bc_latlon()`.

bartnijssen: Not sure why my comment : https://github.com/UW-Hydro/VIC/pull/662/files#r95245324 does not show here, but it should

bartnijssen:
@yixinmao : Please address https://github.com/UW-Hydro/VIC/pull/662/files#r95245324 and I'll merge after that

vic/drivers/shared_image/src/get_global_domain.c
yixinmao: @bartnijssen It shows - and your suggestion sounds good. Will pin you when it's done.

bartnijssen: Weird - it did not show up in the "cover page" earlier .... I blame github

yixinmao:
@bartnijssen I've made the change as you suggested. Should be able to merge.

Issue #663, opened 2017-01-10
VIC 5.0 classic driver stops writing to daily output files after 31 years


tbohn:
Running 5.0, classic mode, hourly time step, daily output step (ndays=1).  The simulation runs to completion (51 years) but the output files end after 31 years (actually, year 31, month 9, day 16, i.e. Sep 15 of the 31st year).

The reason for this appears to be due to a rounding error in the calculation of the hour and dayseconds of the time step.  At the beginning of the simulation, the dayseconds of the first (hourly) step is 0, and the dayseconds of the final (hourly) step is 82800 (=23 hours * 3600 sec/hour).  After about 14 years (year 14, month 9, day 15 of the simulation, with the first year being year 1), the dayseconds of the final step shifts to 86400, and the dayseconds of the first step of the next day shifts to 3600.  This doesn't stop the writing of output though.  Eventually, output files stop being appended to on year 31, month 9, day 16.  This seems to correspond to the dayseconds beginning to look like 82799 instead of 82800, and so on.

I think this is causing the alarm that tells VIC to write its output on the final step of the day to fail.

This is actually creating 2 problems.  First, the alarm is set for dayseconds=82800.  When the first shift occurs, 82800 becomes the dayseconds of hour 22 instead of hour 23, which means that the daily outputs are being written an hour early.  Then, when the dayseconds of hour 22 becomes 82799, the alarm is no longer raised.  So this means that the output files are really only correct for the first 14 years (approximately).

bartnijssen:
This is a critical bug that needs to be resolved before we make the 5.0.1 release since it prevents long runs in classic mode. Anyone willing to take this on?

yixinmao:
@bartnijssen I can do it. Will coordinate with Ted for some sample data.

tbohn:
A little more information on the bug: changing the origin time to 1800-01-01 in vic_time.c didn't help.  However, changing from NDAYS 1 to NSTEPS 24 fixed the problem in terms of getting the output records written for the entire simulation at the correct time of day, except that the values of dayseconds still were miscalculated - and these incorrect values still appeared in the output file record timestamps.

bartnijssen:
Closed per #668

PR #668, opened 2017-01-18
Fix time precision issue for long VIC runs

yixinmao:
Before this fix, the timestamps of long VIC runs were incorrect in some cases. This PR is in response to issue #663.

**Reason**: The reason  of this issue is that during the step where timestamps are made (in `make_dmy`), the length of a timestep is first converted into the specified time unit, denoted as `dt_time_units`; then a series of timestamps for the whole running period is determined by incrementally increase the timestamp by `dt_time_units`. When `dt_time_units` is not precise, the incremented timestamp will accumulate the small error of `dt_time_units`, potentially resulting in incorrect timestamp after a long simulation time.

**Example**: for example, say the time unit is day, and the model timestep is hourly. In this case, `dt_time_units = 1/24`, which is stored as a finite number in the computer with a very small cut-off error. The accumulated error due to this results in an incorrect timestamp shift after ~14 years, and eventually terminates model outputs afeter ~31 years.

**Solution**: when generating the timestamp for each timestep, calculate the accumulating time length from the simulating start time precisely, then store the time. In this way, there is only one cut-off error ofr all timestamps without error accumulation.


yixinmao:
@tbohn , could you take a look at this PR? If you are able to use the new code to test on your data that originally caused error, that'd be great. Thanks!

bartnijssen:
@jhamman : Could you take a quick look at this on travis? The build is failing because of an HDF library issue (is this the same problem someone else mentioned in the past week - it vaguely rings a bell).

jhamman:
The failing tests are all due to an issue with Anaconda, not VIC. I'll look into why this is happening. @yixinmao - did the test suite pass locally?

tbohn:
@yixinmao sure, I'm testing it out now.

jhamman:
@yixinmao - can we get a what's new note?

yixinmao:
@jhamman I'm running the tests locally now (I haven't run the test locally for a while so ran into some issues that are probably not due to this PR...). Do you mean a release note?

jhamman:
yes, a release note.

tbohn:
@yixinmao The code works correctly for me now.  Thanks!

bartnijssen:
Can be merged after travis build passes (issue 669)

PR #681, opened 2017-02-07
Cleaned up vic_store.c the part related to advancing dmy_current for state file timestamp

yixinmao:
In `vic_store()`, we now directly pass in the state time structure (at the state time point) instead of `dmy_current` (which is the beginning of the timestep and one timestep ahead of the state time point). This change will avoid repeatedly advancing one timestep from `dmy_current` in multiple places of the code.

- [x] Image driver
- [x] CESM driver - @dgergel , I think the easiest way is that you pull my branch and make a PR to my branch `clean_up_dmy_vic_store`. Ideally all the remaining changes should happen only under `drivers/cesm`.

### DIFF
--- a/vic/drivers/cesm/src/cesm_interface_c.c
+++ b/vic/drivers/cesm/src/cesm_interface_c.c
@@ -138,6 +138,10 @@ vic_cesm_run(vic_clock *vclock)
     // Write history files
     vic_write_output(&dmy_current);
 
+    // advance the clock
+    advance_time();
+    assert_time_insync(vclock, &dmy_current);
+
     // if save:
     if (vclock->state_flag) {
         // write state file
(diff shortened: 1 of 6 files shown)


bartnijssen:
@yixinmao : Is this ready for review?

yixinmao:
@bartnijssen not yet - still need to finish the according changes in the cesm driver. @dgergel let me know when you get a chance to look at this.

dgergel:
@yixinmao, I've just updated this for the CESM driver. @jhamman would be good if you can take a look; it was a very quick fix that just involved a bit of reordering.

jhamman:
This looks good to me.

yixinmao:
@bartnijssen @jhamman I think this PR is ready to be merged.

PR #231, opened 2015-07-30
Routing extension for VIC image driver.

wietsefranssen:
Affected files in the original VIC files are: `./drivers/image/Makefile`  and `./drivers/image/src/vic_image.c`.

To change the routing scheme once they are implemented, only change the following line in the makefile
-> `ROUT=rout_stub`

Still only an empty routing module is implemented (rout_stub). See ./extensions/rout_stub

closes #24

PR #683, opened 2017-02-16
Fixed some remaining bugs related to reseting forceskip to zero

yixinmao:
In PR #671 that fixed bugs to reset `forceskip` to zero at the beginning of every year, there were some parts missing which are addressed in this PR:
- `forceskip` for forcing[1] (the second forcing file, if specified) should also be reset to zero every year
- The check for new year should start from the second timestep (not the third)


yixinmao:
@bartnijssen @jhamman Actually... should this PR even go into develop branch? Or are we doing another hotfix branch off master?

jhamman:
I think the develop branch sounds like the best option. I want to get #231 closed ASAP so any bug fixes should just get wrapped into 5.1.0 at this point.

PR #688, opened 2017-02-27
Feature/fix compiling issue

dgergel:
- Applies run_cell changes from PR #662 to the CESM driver 
- fixes bug that was occurring when initial moisture was not greater than residual moisture (occurs in the Arctic for a few grid cells at certain time steps when soil moisture is very low), this is fixed by checking to make sure that initial moisture is greater than residual moisture and if not we skip the calc_Q12 calculation, since there should be no drainage in that case. Previously this check was implemented but was not checking the correct variable. 
- addresses issues with a faulty merge (between the debug_CESM_driver and develop branches) that occurred on Thunder (this is the reason for additional files being included in this PR)


dgergel:
Note that the netCDF speedup changes are not included in this PR, so the Travis tests will still fail for the CESM driver. I wanted to put those changes in a separate PR.

PR #692, opened 2017-03-03
Feature/implement runcell changes and netcdf speedup cesm driver

dgergel:
- applies run_cell changes from PR #662 to the CESM driver
- fixes bug that was occurring when initial moisture was not greater than residual moisture (occurs in the Arctic for a few grid cells at certain time steps when soil moisture is very low), this is fixed by checking to make sure that initial moisture is greater than residual moisture and if not we skip the calc_Q12 calculation, since there should be no drainage in that case. Previously this check was implemented but was not checking the correct variable
- applies netCDF speedup changes from #684 to the CESM driver

cc @yixinmao, @jhamman, @bartnijssen for review.


jhamman:
This looks good to me. @yixinmao please give the final review and merge if all tests pass.

PR #697, opened 2017-03-07
Feature/fix aerodynamic resistance and roughness

dgergel:
This PR makes the following changes to the CESM driver: 

- updates the veg hist structure in accordance with the image and classic drivers
- now correctly reads in the FULL_ENERGY option from the global parameter file (it was previously defaulting to False because it wasn't being read in)
- adds printing of global parameter file options after model initialization (this is currently done in image mode and I think it's a very useful feature)

### DIFF
--- a/vic/drivers/cesm/src/cesm_interface_c.c
+++ b/vic/drivers/cesm/src/cesm_interface_c.c
@@ -76,9 +76,6 @@ int
 vic_cesm_init(vic_clock     *vclock,
               case_metadata *cmeta)
 {
-    global_param_struct global_param; 
-    option_struct       options;
-
     // start vic all timer
     timer_start(&(global_timers[TIMER_VIC_ALL]));
     // start vic init timer
--- a/vic/drivers/cesm/src/cesm_interface_c.c
+++ b/vic/drivers/cesm/src/cesm_interface_c.c
@@ -76,6 +76,9 @@ int
 vic_cesm_init(vic_clock     *vclock,
               case_metadata *cmeta)
 {
+    global_param_struct global_param; 
+    option_struct       options;
+
     // start vic all timer
     timer_start(&(global_timers[TIMER_VIC_ALL]));
     // start vic init timer
@@ -101,6 +104,12 @@ vic_cesm_init(vic_clock     *vclock,
 
     // initialize output structures
     vic_init_output(&dmy_current);
+
+    // initialization is complete, print settings
+    log_info(
+	"Initialization is complete, print global param and options structures");
+    print_global_param(&global_param);
+    print_option(&options);
 
     // stop init timer
     timer_stop(&(global_timers[TIMER_VIC_INIT]));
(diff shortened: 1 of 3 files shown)


vic/drivers/cesm/src/cesm_interface_c.c
```diff
@@ -76,6 +76,9 @@ int
 vic_cesm_init(vic_clock     *vclock,
               case_metadata *cmeta)
 {
+    global_param_struct global_param; 
+    option_struct       options;
+
```
jhamman: I don't think you want to / need to declare these here. They are in the header of this file.

dgergel: Oops, missed that. I removed them.

vic/drivers/cesm/src/get_global_param.c
```diff
@@ -72,6 +72,10 @@ get_global_param(FILE *gp)
                 sscanf(cmdstr, "%*s %s", flgstr);
                 global_param.time_units = str_to_timeunits(flgstr);
             }
+	    else if (strcasecmp("FULL_ENERGY", optstr) == 0) {
+		sscanf(cmdstr, "%*s %s", flgstr);
+		options.FULL_ENERGY = str_to_bool(flgstr);
+	    }
```
bartnijssen: I think that it was not read in as part of the CESM driver, because coupled model simulations ONLY make sense with `options.FULL_ENERGY` set to `TRUE`. However, you could imagine a water balance run with a data atmosphere, so this would be OK. Since that is a special case for the CESM driver, it would be good to print out a warning if `options.FULL_ENERGY` is set to `FALSE`.

So I suggest changing:
```C
            else if (strcasecmp("FULL_ENERGY", optstr) == 0) {
                sscanf(cmdstr, "%*s %s", flgstr);
                options.FULL_ENERGY = str_to_bool(flgstr);
            }
```
to
``` C
            else if (strcasecmp("FULL_ENERGY", optstr) == 0) {
                sscanf(cmdstr, "%*s %s", flgstr);
                options.FULL_ENERGY = str_to_bool(flgstr);
                if (options.FULL_ENERGY == FALSE) {
                    log_warn("FULL_ENERGY is set to FALSE. Make sure that this is intentional.");
                }
            }
```

bartnijssen:
Please use some of the `print_<struct>` to check that all the structures are populated with the data you expect from the parameter files. You should check this for both the vegetation and soil data. These would just be debugging additions.

vic/drivers/cesm/src/get_global_param.c
dgergel: Good point, I've updated this.

vic/drivers/cesm/src/get_global_param.c
```diff
@@ -72,6 +72,14 @@ get_global_param(FILE *gp)
                 sscanf(cmdstr, "%*s %s", flgstr);
                 global_param.time_units = str_to_timeunits(flgstr);
             }
+	    else if (strcasecmp("FULL_ENERGY", optstr) == 0) {
+		sscanf(cmdstr, "%*s %s", flgstr);
+		options.FULL_ENERGY = str_to_bool(flgstr);
+		if (options.FULL_ENERGY == FALSE) {
```
jhamman: `FALSE` --> `false`

dgergel: oops, and that's why the build was failing. updated.

PR #699, opened 2017-03-08
feature/fix_vapor_pressure_deficit_check

dgergel:
This PR implements a check on vapor pressure deficit in the CESM driver (if VPD is < 0, it sets VPD to zero and updates VP) as is currently implemented in the image driver.

### DIFF
--- a/vic/drivers/cesm/src/vic_force.c
+++ b/vic/drivers/cesm/src/vic_force.c
@@ -206,6 +206,10 @@ vic_force(void)
             // vapor pressure deficit
             force[i].vpd[j] = svp(force[i].air_temp[j]) - force[i].vp[j];
 	    if (force[i].vpd[j] < 0) {
+		log_warn("Vapor pressure deficit is %f which is < 0, "
+			 "setting vapor pressure deficit to 0 and calculating "
+			 "saturated vapor pressure using air temperature %f.",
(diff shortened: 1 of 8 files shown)


jhamman:
See also https://github.com/UW-Hydro/VIC/pull/573.

bartnijssen:
In a coupled model this should not happen, but good to have the check. However, I share Joe's concern about the potential for a (tiny) leak, so a `log_warn()` would be in order. The user can always turn those off by changing the `LOG_LEVEL`.

vic/drivers/cesm/src/vic_force.c
```diff
@@ -205,6 +205,10 @@ vic_force(void)
         for (j = 0; j < NF; j++) {
             // vapor pressure deficit
             force[i].vpd[j] = svp(force[i].air_temp[j]) - force[i].vp[j];
+	    if (force[i].vpd[j] < 0) {
+		force[i].vpd[j] = 0;
+		force[i].vp[j] = svp(force[i].air_temp[j]);
+	    }
```
bartnijssen: Add `log_warn()` when this happens with info about the reset (display the value of `vpd` before the reset and say it was rest to 0).

dgergel:
That makes sense - I've updated this to include a warning.

PR #702, opened 2017-03-16
fix julian day in dmy struct for first time step in CESM driver

### LINKED ISSUE #700: fix Julian day in CESM driver time structure
Currently, the Julian day in the vic time structure (`dmy_struct`) is not being set correctly - for a start date of September 1, it is 0, rather than 245. This is because it is not being set in `vic_cesm_time.c` in the CESM driver, so it defaults to 0. We need to set it directly in the `dmy_struct` to solve this problem.

dgergel:
- closes #700 
- fixes Julian day in CESM driver (was previously not being set correctly for the first time step)


jhamman:
Looks good. Add a note to the change log and then you can merge.

PR #713, opened 2017-04-24
Feature/fix memory leaks

dgergel:
This PR fixes a number of memory leaks that were identified using `valgrind`.


bartnijssen:
@dgergel : I'll review after the checks pass - they currently fail for all builds (which means it's probably something straightforward)

vic/drivers/shared_image/src/vic_store.c
```diff
@@ -1266,6 +1266,9 @@ vic_store(dmy_struct *dmy_state,
 
     free(ivar);
     free(dvar);
+
+    // free memory
```
bartnijssen: remove whiteline (1269) and comment (1270) - the previous lines also free memory so this is nothing different

dgergel:
@bartnijssen I've addressed your comments, so this is ready to merge.

PR #716, opened 2017-06-07
Feature/fix initialization of coupler fields

dgergel:
- [ ] ~~closes #xxx~~
- [X] tests passed
- [X] ~~new tests added~~
- [ ] ~~science test figures~~
- [X] ReleaseNotes entry

This PR does the following: 
- sets the temperature field in the `l2x_data_struct` (`l2x_Sl_t`) during initialization using `avg_temp` from the `soil_con_struct`, which is equivalent to using the temperature of the top soil layer, since this is the value we use to initialize soil temperatures from a cold start

- sets the upwelling longwave field in the `l2x_data_struct` (`l2x_Fall_lwup`) using `l2x_Sl_t` and emissivity from the `param` struct


dgergel:
Note: we may want to combine `vic_initialize_albedo`, `vic_initialize_temperature` and `vic_initialize_lwup` into one function.

vic/drivers/cesm/include/vic_driver_cesm.h
```diff
@@ -36,6 +36,7 @@
 #define GLOBALPARAM "vic.globalconfig.txt"
 #define RPOINTER "rpointer.lnd"
 #define SHR_CONST_SPVAL 1.0e30   /**< CESM missing value */
+#define SHR_CONST_STEBOL 5.67e-8 /**< Stefan-Boltzmann constant ~ W/m^2/K^4 */
```
jhamman: this line/variable should be removed and we should use the equivalent constant in `vic_run`.

dgergel: Updated.

vic/drivers/cesm/src/vic_cesm_init_library.c
```diff
@@ -88,37 +88,37 @@ initialize_l2x_data(void)
...
+        l2x_vic[i].l2x_Fall_evap = 0;
+        l2x_vic[i].l2x_Fall_swnet = 0;
+        l2x_vic[i].l2x_Fall_fco2_lnd = 0;
+        l2x_vic[i].l2x_Fall_flxdst1 = 0;
+        l2x_vic[i].l2x_Fall_flxdst2 = 0;
+        l2x_vic[i].l2x_Fall_flxdst3 = 0;
+        l2x_vic[i].l2x_Fall_flxdst4 = 0;
+        l2x_vic[i].l2x_Fall_flxvoc = 0;
+        l2x_vic[i].l2x_Flrl_rofliq = 0;
+        l2x_vic[i].l2x_Flrl_rofice = 0;
```
dgergel: I've gone back to setting all to `SHR_CONST_SPVAL` for now. But I think we should revisit this, as we set the coupler fields to 0 previously for VIC in RASM (I'll make an issue for it so we can figure out what we want to do).

PR #726, opened 2017-08-07
fix evap units and sign

### LINKED ISSUE #727: fix units and sign for evap passed to coupler CESM driver
Currently we are converting `evap` to m by multiplying by `MM_PER_M`, which is not necessary to convert mm to kg m2/s, before we pass it to the coupler in the CESM driver. We're also multiplying by a negative sign. Neither of these are necessary - we just need to divide by `global_param.dt`.

dgergel:
- [X] closes #727 
- [X] tests passed
- [X] new tests added
- [X] science test figures
- [X] ReleaseNotes entry

### DIFF
--- a/vic/drivers/cesm/src/cesm_put_data.c
+++ b/vic/drivers/cesm/src/cesm_put_data.c
@@ -158,7 +158,6 @@ vic_cesm_put_data()
         l2x_vic[i].l2x_Fall_sen += -1 * out_data[i][OUT_SENSIBLE][0];
 
         // evaporation, VIC: mm, CESM: kg m-2 s-1
-        // TO-DO should we incorporate bare soil evap?
         l2x_vic[i].l2x_Fall_evap += out_data[i][OUT_EVAP][0] /
                                     global_param.dt;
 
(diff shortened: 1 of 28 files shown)


bartnijssen:
Which #xxx is this closing (the point is to indicate which issue this PR is related to)?

vic/drivers/cesm/src/cesm_put_data.c
```diff
@@ -159,9 +159,8 @@ vic_cesm_put_data()
 
         // evaporation, VIC: mm, CESM: kg m-2 s-1
         // TO-DO should we incorporate bare soil evap?
```
bartnijssen: Is this TODO still relevant? My assumption is that WRF wants total evap, so if soil evap is already included in `OUT_EVAP` then we should be OK. If not, then we should add it.

dgergel: It's not still relevant. Since we're using `evap` from `OUT_DATA` here, it's already incorporating soil evap.

dgergel:
@bartnijssen I hadn't opened an issue. Will do that so that we can close it with this PR.

PR #734, opened 2017-08-18
Feature/fix remaining coupler fields

dgergel:
This PR fixes the signs for evaporation (`l2x_Fall_evap`), latent heat (`l2x_Fall_lat`) and sensible heat (`l2x_Fall_sen`). Sign conventions now agree with those for coupled models (before we'd switched them back to agree with the image driver, which was incorrect). 

- [ ] ~closes #xxx~
- [x] tests passed
- [ ] ~new tests added~
- [ ] ~science test figures~
- [X] ReleaseNotes entry

### DIFF
--- a/docs/Development/ReleaseNotes.md
+++ b/docs/Development/ReleaseNotes.md
@@ -69,7 +69,11 @@ To check which release of VIC you are running:
 
     [GH#732] (https://github.com/UW-Hydro/VIC/pull/732)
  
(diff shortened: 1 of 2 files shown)


dgergel:
Just a few figures to show the turbulent heat flux changes and evaporation changes; these figures are from an RBR run that ran until the debug queue walltime was up, so the figures are from the last timestep before that (from Sept 1 1979). Colorbars are not constrained to show the full ranges for both. 

image

image

image

PR #742, opened 2017-08-25
Fix/issue673

### LINKED ISSUE #673: Fix to typo in image driver vic_force.c
VIC 5.0.0 image driver - OS, compiler, settings, etc., are unimportant

In vic_force.c, for the "fcan" variable, the code expects the forcing file to name this variable "fcov".  This needs to be changed to "fcan" to match VIC's name for the variable.

(In fact, the reason this instance of "fcov" exists is due to an incomplete conversion of the variable name from "fcov" to "fcan" throughout the code, which I was responsible for - this is simply a mistake.)

The fix for this should be extremely minor and will have no effects on existing tests or performance since this feature is not yet commonly used.

bartnijssen:
- [x] closes #673
- [x] tests passed
~- [ ] new tests added~
~- [ ] science test figures~
~- [ ] ran uncrustify prior to final commit~
- [x] ReleaseNotes entry

### DIFF
--- a/docs/Development/ReleaseNotes.md
+++ b/docs/Development/ReleaseNotes.md
@@ -99,9 +99,11 @@ To check which release of VIC you are running:
 
 	The VIC image and CESM drivers now may be optionally compiled with OPENMP to enable shared memory thread parallelization. This option should improve the parallel scaling of these drivers by reducing the number of MPI messages and increasing message size.
 
-	8. Cleanup
+#### Bug Fixes:
 
-	([GH#673](https://github.com/UW-Hydro/VIC/pull/673)) Renamed "fcov" to "fcan" in image driver to better match variable code name.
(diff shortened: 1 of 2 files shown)


docs/Development/ReleaseNotes.md
```diff
@@ -99,6 +99,10 @@ To check which release of VIC you are running:
 
 	The VIC image and CESM drivers now may be optionally compiled with OPENMP to enable shared memory thread parallelization. This option should improve the parallel scaling of these drivers by reducing the number of MPI messages and increasing message size.
 
+	8. Cleanup
```
jhamman: this is indented too far.

jhamman: You probably should put this in the bug fixes section too.

PR #751, opened 2017-10-20
Small final updates for the routing extension, including:

yixinmao:
- Minor docs update
- Rename extension functions to be routing-extension-specific


wietsefranssen:
@jhamman @yixinmao We agree with your suggestions, however bear in mind that in case another (not routing) extension is added in the future, this naming convention may cause inconsistencies.

PR #771, opened 2018-01-16
Feature/move max iter grnd canopy

### LINKED ISSUE #770: move MAX_ITER_GRND_CANOPY to parameters struct
Currently, to modify the number of nested ground-canopy iterations in the `surface_fluxes.c` routine in `vic_run`, `MAX_ITER_GRND_CANOPY` has to be altered within the `surface_fluxes.c` routine. It could be moved to the parameters struct so that it can be altered more easily in the event of wanting to reduce the number of maximum iterations before falling back.

dgergel:
- [X] closes #770 
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [X] ReleaseNotes entry


vic/vic_run/src/surface_fluxes.c
```diff
@@ -225,11 +224,9 @@ surface_fluxes(bool                 overstory,
...
     double            store_NPP;
 
-    if (options.CLOSE_ENERGY) {
-        MAX_ITER_GRND_CANOPY = 10;
-    }
-    else {
-        MAX_ITER_GRND_CANOPY = 0;
+    if (!options.CLOSE_ENERGY) {
+        // if not closing energy balance, don't iterate
+        param.MAX_ITER_GRND_CANOPY = 0;
```
jhamman: can we move this change to the parameters initialization? No need to reset this variable each timestep.

dgergel: I considered doing that but moving it to `vic/drivers/shared_all/src/initialize_parameters.c` requires using the `options` struct in that routine, which we don't currently do - do you think that's fine?

jhamman: yes, I think that it probably cleanest.

dgergel: I've updated this as you suggested - I'll ping you once the tests pass and it's ready for another review.

dgergel:
@jhamman this is ready for another review.

PR #776, opened 2018-02-01
Feature/fix deltacc bug

### LINKED ISSUE #775: extremely large OUT_ENERGY_ERROR values occur when running VIC 5 over RASM domain
When running VIC 5 over the RASM domain for a year, outputting `OUT_ENERGY_ERROR` results in unreasonably large values, on the order of `10^12`. Further investigation of VIC 5 behavior shows that the extremely large energy error values are coming from an unreasonably large (and negative) `deltaCC` term - the term representing change in cold content of the snowpack.

dgergel:
- [X] closes #775 
- [x] tests passed
- [ ] new tests added
- [ ] science test figures
- [X] ReleaseNotes entry

This PR fixes the issue described in #775. In addition to what is described there, the unreasonable `deltaCC` values were a result of incorrect values for the ground flux out of the snowpack, which is subtracted from the `deltaCC` term. Calculating the ground flux out of the snowpack depends on snow density (squared) and snow depth, and extremely large values for snow density were occurring for certain grid cells at certain time steps in the RASM domain. This was fixed by a) implementing a cap on the new snow density calculated by the Hedstrom and Pomeroy 1998 equation, and b) disregarding the ground flux out of the snowpack if snow depth is below a certain threshold (`1.e-8`). A new parameter, `SNOW_NEW_SNOW_DENS_MAX`, was added to the parameters struct to cap snow density.


bartnijssen:
Looks good to me : @jhamman - another pair of eyes may be useful. Feel free to merge.

vic/vic_run/src/SnowPackEnergyBalance.c
```diff
@@ -105,6 +109,7 @@ SnowPackEnergyBalance(double  TSurf,
                                      intercepted snow (kg/m2s) */
     double BlowingMassFlux;       /* Mass flux of water vapor from blowing snow. (kg/m2s) */
     double SurfaceMassFlux;       /* Mass flux of water vapor from pack snow. (kg/m2s) */
+    double epsilon_snow = 1.e-8;  /* Snow depth below which we ignore the ground flux from the snowpack */
```
jhamman: This should be a parameter in the parameters structure.

dgergel:
@jhamman - I've updated this to add the snow depth threshold as a parameter rather than a hard coded literal. Ready for another review

PR #780, opened 2018-02-16
Feature/fix close energy bug

### LINKED ISSUE #779: Large difference in vic run walltime before and after vic_run cleanup and veg classes memory allocation changes
Assessing VIC RUN walltime between January 17 and January 18 (before and after the `vic_run` and `prepare_full_energy` changes were made shows large timing differences over the RASM domain. Before the PR was merged, maximum walltime for a single grid cell in the RASM domain over a year-long period was up to 2.5s, but after the PR was merged maximum run time decreased to 0.2 s over the entire domain and time period for a single timestep. 

Looking more closely at the actual values, and run for a single grid cell in the RASM domain (located in Central Asia), SWE is different before and after the Jan 17th change - this is a plot showing SWE from Sept 1948 - Aug. 1949 with the VIC 5 image driver run at a 3-hourly timestep for the gridcell in the RASM domain in Central Asia, showing results from: a) pre vic run change, b) post vic run change, and c) delta cc bug fix (merged into `develop` after the vic run cleanup PR from Jan 17)

image

We need to investigate these changes further and look at the science test results to determine what's going on. 

cc @tbohn @jhamman @yixinmao @bartnijssen

dgergel:
…rs and then update in surface_fluxes

This PR fixes a bug that was introduced in #771 (`MAX_ITER_GRND_CANOPY` was not being set when `CLOSE_ENERGY` was set to `true`, thus there were no iterations). 

- [X] closes #779
- [x] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry


vic/vic_run/src/vic_run.c
```diff
@@ -131,6 +131,13 @@ vic_run(force_data_struct   *force,
         gauge_correction[0] = 1;
         gauge_correction[1] = 1;
     }
+
+    // Canopy Iterations
+    if (!options.CLOSE_ENERGY) {
+        // do not iterate to close energy balance
+        param.MAX_ITER_GRND_CANOPY = 0;
+    }
+
```
jhamman: I still don't like this. I get why this can't be in `initialize_parameters.c` but as a general rule, we shouldn't be modifying the parameters structure inside `vic_run`. Can you find a better place for this inside `vic_init`?

dgergel: I agree, I don't like it either but I couldn't think of a better option. I hadn't thought of putting it in `vic_init` since we aren't changing any parameter values in there but I guess we could. 

@jhamman what do you think about only updating (as currently in `vic_run`) when we are not running with `CLOSE_ENERGY` being set to true? I actually don't like that but I can't think of a better way to do it since we've made `MAX_ITER_GRND_CANOPY` a parameter and we don't want to hard code it except in the case of `CLOSE_ENERGY` being false

jhamman: I don't recall the full order of operations but you there should be a straightforward way to do this somewhere within the `vic_init` space. It just needs to be updated after both values are available. Can you look in `vic_init` and tell me at what point they are both populated?

dgergel: I've already updated `vic_init` and am testing it out right now - stand by

dgergel:
@jhamman this is ready for another review

dgergel:
@bartnijssen @jhamman it would be great if we could get this merged soon since it's is a bug fix for `CLOSE_ENERGY`. It's ready for a final review.

Issue #790, opened 2018-05-12
multiple definition error when referenced by c++


Sibada:
- VIC version 5.0.1 February 1, 2017
- gcc 4.9.3; g++ 4.9.3
- MinGW of RBuildTools 3.3

#### Feature Requests
I try to write an c++ extension of VIC to make it could be called by R easier. I found that some global variables were definited in the head files, including `char vic_run_ref_str[MAXSTRING];` at line 95 of file `vic/vic_run/include/vic_def.h`, `FILE *LOG_DEST;` at line 61 of file `vic/vic_run/include/vic_log.h`, and `double (*funcd)(double z, ...` at line 177 of file `vic/vic_run/include/vic_run.h`.
It causes multiple definition error when I let several cpp files include the `vic_run.h`.
It seems to solved when I take them into a c file and place `extern` to their original definitions. Might it be proper?

jhamman:
@Sibada - thank you for the report. 

To be honest, we don't have much experience with C++. In plain C, we protect includes with 

```C
#ifndef VIC_RUN_H
#define VIC_RUN_H
...
#endif
````

Does this not work with C++?

Before we make any changes to VIC to support C++, we'd want to get Travis setup to also test with a C++ compiler. I know @mschnorb and @jameshiebert also have interest in using VIC with C++ so its not a terrible idea to explore this with VIC.5.

-----

> extension of VIC to make it could be called by R easier 

I recognize there may be some interest here as I've heard about this from others as well. Perhaps it would be worth opening a separate issue to discuss how that would work.

jameshiebert:
I'm not a C++ expert either, but IIRC we moved to C++ on VIC in our fork off of VIC 4.0.7 without much fuss using g++. I'd experiment with your compiler flags before you start ripping up code.

Sibada:
@jhamman Yeah, it seems that C++ do not support placing global parameters in the head file even the `#ifndef`  et al. were used, and I found that many other C and C++ developers suggest to place global parameters in a source file and used `extern` to define them in the head files.

jhamman:
@Sibada - I think I understand. It may be instructive to see which changes you made to support using VIC with a C++ compiler. Would it be possible for you to open a PR with just those changes? Even better if you can enable a travis build with a C++ compiler too.

Sibada:
@jhamman Those revises are applied on the master branch, might I should revise them on the develop branch and open the PR?
My revises includes add an `extern` to the h files:
```C
[...]

PR #792, opened 2018-05-19
Changes discussed in #790

Sibada:
- [ ] closes #xxx
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry

There are failure in the unit test:
```bash
=================================== FAILURES ===================================
_________________________________ test_strpdmy _________________________________
def test_strpdmy():
        dmy = ffi.new("dmy_struct *")
    
        dates = pd.date_range('2015-12-18', '2016-12-22', freq='1D')
        date_format = '%Y-%m-%d'
        for date in dates:
            date_str = date.strftime(date_format)
            vic_lib.strpdmy(date_str.encode(), date_format.encode(), dmy)
>           expected = date.to_datetime()
E           AttributeError: 'Timestamp' object has no attribute 'to_datetime'
tests/unit/shared/test_vic_time.py:374: AttributeError
========== 1 failed, 55 passed, 16 skipped, 1 xpassed in 4.80 seconds ==========
```
But the tests of examples seem no failure. Might it because of my changes?


jhamman:
@Sibada - We're seeing a bunch of test failures here with the following message: 

> cffi.error.FFIError: multiple declarations of variable LOG_DEST (for interactive usage, try cdef(xx, override=True))

Any thoughts on how to fix this?

jhamman:
To answer your above question, the tests are passing on develop so yes, I suspect your changes are responsible for the failures.

jhamman:
@Sibada - the 'Timestamp failure has been fixed in develop. There are still failures here to address. 

Something you may consider doing is enabling a c++ build of VIC on Travis and maybe even setting up Appveyor to test VIC using Windows. I think those changes are the logical first steps in supporting C++/windows with VIC. 

Travis C++: https://docs.travis-ci.com/user/languages/cpp/
Appveyor C++: https://www.appveyor.com/docs/lang/cpp/

Sibada:
Feel sorry for no replying for these days because I'm busy for revise of my paper and development of the R package. The package finally can run on my win7 platform computer.
Might `multiple declarations of variable LOG_DEST` is because no `extern` in some source files?

Sibada:
I'm sorry for my no replying because I'm busy in revising my paper and development of the R package. It seems successfully compiled and run on my R platform on my win7 computer.

Sibada:
After the development of the R package, I found that little modification of VIC source code such as in this puul request make little sense because of large difference between the raw terminal environment of os and the R runtime environment. For example, in the original VIC source code `exit()` is used to stop the running when error occurs but when it is used in R, it would terminate the entire R process and might cause irretrievable losses, therefore a specific exit function for R are expected. Several places of the original source code are also needs to be modified when adopted to R due to the strict policy of CRAN. In fact, to pass the R CMD check (determined whether this R package could be technologically accepted by CRAN), the original VIC source code I placed in the R package have been beyond recogniton.

Sibada:
@jhamman
In my view, it might be less worthy to make individual drivers use their own exit functions. To realize that we can move the definition of `log_err()` and `log_warn()` to a source file determined each drivers but it would cause higher coupling between the modules.
I also found that the R environment also needs specified print function. The original `printf()` of C also can work in R environment but CRAN could not accept it. In fact I could not exactly remember what I have done in the VIC source codes...
However I have some suggestions for the VIC source codes and it might be useful:
I found that the codes that check or transform the parameters after they're read in (e.g. those calculate soil properties,  critical and wilting point moisture contents of soil after the parameters have been get from file and assigned to the soil_con struct) are usually just placed after the codes that read the parameters, and they are repetitively in different drivers (classic and image). I thought it might be more proper to set them as a single function in the share_all module and it can be used by several drivers.
In some functions that transform parameter from string type to enum (e.g. str_to_XXX), the arguement of the input string (`char* xxx`) might be better to use (`const char* xxx`). The string objects of C++ can provide their char* but it is const.

My package have not upload to CRAN yet because I have to deal with how to pass the R CMD Check. For linux platform there are no ERROR and WARNING but some WARNINGs exist in windows platform.
Can you have a try to this package and give some suggestions? I place it in a China netdisk because most US netdisks are blocked by the GFW.
https://pan.baidu.com/s/10N6gcbzU6DyoAyK_Wmytew
And can I use your STEHE dataset as sample dataset for this package?

jhamman:
@Sibada - thank you for your detailed response. Happy to keep this closed and we can revisit down the road if/when others are interested.

PR #794, opened 2018-05-21
Fixed incorrect apportioning of root fractions.

### LINKED ISSUE #793: [Bug] Incorrect root fraction calculation
- VIC version: 5.1.0.rc1
- Problem Description:
-- In some cases, `calc_root_fraction()` computes substantially different root fractions from those of prior releases. This leads to substantially different rates of transpiration.
-- This change arose in PR #701, which included some general cleanup fixes, including a fix to `calc_root_fraction.c` that was intended to fix problems with occasional cases of infinite loops in the algorithm.  The fix did eliminate infinite loops, but changed the apportioning of root fractions to soil layers. One line in particular seems to have caused the change in apportioning: `zone ++;` was inserted on line 137.
- Proposed Solution:
-- Fix the algorithm to apportion correctly and also end the loop correctly.

tbohn:
- [x] closes #793
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry

This actually is a follow-up to PR #701, in which we made a change to `calc_root_fraction.c` to fix a problem with infinite loops, but in the process ended up incorrectly apportioning the root fractions to the soil layers. Over parts of the Colorado Basin, I have observed VIC 5.1 producing substantially different transpiration from VIC 4.2, and I've been able to determine that most of the differences are due to the new root zone apportionment.

My new code here is a new algorithm for computing the fractions that doesn't involve any calls to `linear_interp()`.  The approach is to break up the soil column into depth intervals `dD` that span from the previous soil layer OR root zone boundary to the next; i.e. it is the set of spaces between the union of soil layer boundaries and root zone boundaries.  For each soil layer, we compute a weighted sum of root fractions over the various `dD`s that occur in the soil layer.  To accomplish the weighted sum efficiently, I first compute the root fraction density in each zone: `root_dens[zone] = fract[zone] / depth[zone]`.  This density is then an "intrinsic" variable that can be easily piece-wise integrated over a soil layer.

So far, I've tested this algorithm in 2 ways:
(1) within a python notebook, I've ensured that the algorithm works for various cases, including where the bottom of the lowest root zone is either deeper than or shallower than the bottom of the soil column.
(2) within VIC, I've tested this over the San Juan sub-basin of the Colorado Basin and ensured that it reproduced the root fractions (and fluxes) that VIC 4.2 gave.


vic/drivers/shared_all/src/calc_root_fraction.c
```diff
@@ -41,126 +41,121 @@ calc_root_fractions(veg_con_struct  *veg_con,
...
+                     Dsum == Zsum ) ) ) {
+                ltmp = layer;
+                if (layer >= options.Nlayer - 1) {
+                    ltmp = options.Nlayer - 1;
                 }
-                if (Zsum + Zstep <= Lsum) {
-                    /** Root zone ends in current layer **/
-                    Zmax = Zsum + Zstep;
+                if (sum_depth > 0) {
+                    veg_con[veg].root[ltmp] = dL * sum_dens / sum_depth;
```
yixinmao: It seems like `sum_dens` already accumulates the fraction in one sol layer, so why do we need to apply further rescaling here? (After chatting with Ted offline we think `sum_dens` and `sum_depth` may be the same in most cases, and the code could be simplified)

tbohn: Correction: we think `sum_depth` and `dL` may be the same - and might therefore always cancel out.  I originally set up the code this way because I thought I'd need to allow `sum_depth` and `dL` to be unequal, e.g., if the bottom of the soil column and the bottom of the root zones don't match, but the way the code ended up, I think `Lsum` and `Zsum` end up being set equal to each other in these cases, meaning that `sum_depth` and `dL` would end up being equal all the time.  I'm going to test out a simplification of the code and upload it if it works.  Stay tuned...

vic/drivers/shared_all/src/calc_root_fraction.c
```diff
@@ -41,128 +41,126 @@ calc_root_fractions(veg_con_struct  *veg_con,
...
+            // Wait to do this until either we've completed a soil layer
+            // (Lsum > Lsumprev && Dsum == Lsum)
+            // or we're at the final layer and we've completed all root zones
+            // (layer >= options.Nlayer - 1 &&
+            //  zone >= options.ROOT_ZONES - 1 && Dsum == Zsum)
+            if ( Lsum > Lsumprev && Dsum == Lsum &&
+                 ( layer < options.Nlayer - 1 ||
+                   ( layer >= options.Nlayer - 1 &&
+                     zone >= options.ROOT_ZONES - 1 &&
+                     Dsum == Zsum ) ) ) {
```
yixinmao: There seems to be a parenthesis problem - should it be: `( Lsum > Lsumprev && Dsum == Lsum && ( layer < options.Nlayer - 1 ) || ( layer >= options.Nlayer - 1 && zone >= options.ROOT_ZONES - 1 && Dsum == Zsum ) ) `? (note the parenthesis after `layer < options.Nlayer -1`) Double check

vic/drivers/shared_all/src/calc_root_fraction.c
```diff
@@ -41,128 +41,126 @@ calc_root_fractions(veg_con_struct  *veg_con,
...
 
-        if (sum_fract > 0 && layer >= options.Nlayer) {
-            veg_con[veg].root[options.Nlayer - 1] += sum_fract;
-        }
-        else if (sum_fract > 0) {
-            veg_con[veg].root[layer] += sum_fract;
         }
 
+        // Final check on root fractions. If they don't sum to 1, throw error
+        // Otherwise, rescale by sum to eliminate small rounding errors
```
yixinmao: I think it would be a good idea to also add a check to see whether the sum of `veg_con[veg].root[layer]` is close to one before rescaling (and throw an error if not), to prevent input/algorithm error

vic/drivers/shared_all/src/calc_root_fraction.c
```diff
@@ -41,128 +41,126 @@ calc_root_fractions(veg_con_struct  *veg_con,
...
-        sum_fract = 0;
+
+        // Compute density of root fractions in each root zone
+        for (zone = 0; zone < options.ROOT_ZONES; zone++) {
+            if (veg_con[veg].zone_depth[zone] > 0) {
+                root_dens[zone] = veg_con[veg].zone_fract[zone] /
+                                  veg_con[veg].zone_depth[zone];
+            }
+            else {
+                root_dens[zone] = 1.0;
```
yixinmao: Is there a reason why setting `root_dens[zone] = 1.0` in the case that `zone_depth[zone] == 0`? (I guess in practice, no matter what value `root_dens[zone]` is, it will eventually be multiplied by zero so it doesn't really matter)

vic/drivers/shared_all/src/calc_root_fraction.c
tbohn: No, the parentheses are the way I intended them...  But I suppose I could remove the `layer >= options.Nlayer -1 condition in the second set of parentheses, since it's redundant...

vic/drivers/shared_all/src/calc_root_fraction.c
tbohn: Exactly, the value doesn't matter.  I like `1.0` because if you ever want to take the inverse, it is defined.

vic/drivers/shared_all/src/calc_root_fraction.c
tbohn: OK, will do.

vic/drivers/shared_all/src/calc_root_fraction.c
```diff
@@ -41,128 +41,126 @@ calc_root_fractions(veg_con_struct  *veg_con,
...
-    double               Zmin_depth;
-    double               Zmax;
+
+    root_dens = calloc(options.ROOT_ZONES, sizeof(double));
 
     Nveg = veg_con[0].vegetat_type_num;
 
     for (veg = 0; veg < Nveg; veg++) {
-        sum_fract = 0;
+
```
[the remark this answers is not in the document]
tbohn: OK, will do.

vic/drivers/shared_all/src/calc_root_fraction.c
```diff
@@ -41,141 +54,157 @@ calc_root_fractions(veg_con_struct  *veg_con,
...
+        // Check that root fractions sum to 1.0
+        dum = 0.0;
+        for (zone = 0; zone < options.ROOT_ZONES; zone++) {
+            dum += veg_con[veg].zone_fract[zone];
+        }
+        if (fabs(dum - 1.0) > 1.e-4) {
```
jhamman: suggest using `assert_close_double(dum, 1, 0, 1e-4)` here.

vic/drivers/shared_all/src/calc_root_fraction.c
```diff
@@ -41,141 +54,157 @@ calc_root_fractions(veg_con_struct  *veg_con,
...
         for (layer = 0; layer < options.Nlayer; layer++) {
             if (veg_con[veg].root[layer] < 1.e-4) {
                 veg_con[veg].root[layer] = 0.;
             }
             dum += veg_con[veg].root[layer];
         }
-        if (dum == 0.0) {
-            log_err("Root fractions sum equals zero: %f , Vege Class: %d",
-                    dum, veg_con[veg].veg_class);
+        if (fabs(dum - 1.0) > 1.e-4) {
```
jhamman: use `assert_close_double`

jhamman:
(btw, I just updated your branch with my unit test fix so you'll need to update your local copy before making changes).

tbohn:
I'm evaluating runoff over the San Juan basin with various combinations of fixes; will let you know the end result soon.

tbohn:
Here are the results over the San Juan, for runoff+baseflow (long-term average flux in m3/s), 2005-2010:
 - 4.2: 86.8
 - 5.1: 70.9
 - 5.1.fix_calc_root_fraction: 88.3
 - 5.1.fix_calc_root_fraction_and_rollback_LW_EMISS_commit: 86.2
So, as you can see, the various changes between 4.2 and 5.1, including the bug in `calc_root_fraction()` caused an 18% reduction in channel flow.  Fixing this bug restored the losses, exposing a new difference of +2%.  I traced this difference back to 2 changes that impact SWE: (1) the introduction of an emissivity < 1.0 in the outgoing longwave calculation (`calc_out_longwave()`); and (2) the cap on snow density, which impacts ground flux out of the snow pack.  Both of those changes are improvements and should be kept, but they do result in faster melt rates from ground and canopy snow packs, and could therefore invalidate calibrations in some places.  I also looked at the impact of the more accurate Q12 formulation in `runoff()`, but its impact was restricted to brief instances of very wet soil.

jhamman:
Okay, I'm a believer. @bartnijssen - can you give this a look-see?

PR #796, opened 2018-05-22
update to include documentation on running uncrustify

### LINKED ISSUE #783: Documentation needs to list the version of uncrustify that is required for development
- VIC Version: 5.0
- Compiler: gcc 5.4.0
- OS: ubuntu 
- Problem: running uncrustify is a required step when developers submit PRs.  However, running the wrong version of uncrustify will result in undesirable code changes.  On Ubuntu at least, the latest version of uncrustify currently available for a simple installation (v 0.59) is too old, and gives bad results.  Upgrading to 0.64 or later fixes the problem (we don't know the exact version that fixes the problem; all we know is that 0.64 is new enough).  This required extra effort on my part to explicitly download the source code for a newer version of uncrustify and build it.
- Proposed solution: we should add a clear statement to the online documentation that uncrustify 0.64 or later is required, if one wants to submit a PR (and fulfill the uncrustify requirement of the PR).  It's only necessary for developers, so I guess the statement should go somewhere in the development section.
- Note also that running uncrustify within the vic_test_env environment MIGHT also be necessary.  Haven't tested that for sure.  I'll try to find out and update this.

dgergel:
- [X] closes #783 
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry


docs/Development/working-with-git.md
```diff
@@ -128,6 +128,9 @@ To register the deletion of a file:
...
 
+### 3. Correcting code syntax issues prior to creating a PR
+After completing your code commits, you also need to check that your code is compliant with C conventions for indentation and spacing. Please make sure you have `uncrustify` installed, and you will need either version `0.64` or later. Once you have installed `uncrustify`, you should navigate to the `~/tools/code_format` directory in the VIC repo and run `uncrustify`, which you can do by running the script in that directory, `./run_uncrustify.bash`. A simple `git diff --name-only` will list all files that `uncrustify` has updated the syntax for. If any files come up that you have edited, you should commit those syntax edits.
+
```
jhamman: I've made a few changes to the paragraph...

After completing your code changes, you also need to check that your code is compliant with VIC's style conventions for indentation and spacing. Please make sure you have `uncrustify` installed (version `0.64` or later). Once you have installed `uncrustify`, you should navigate to the `...VIC/tools/code_format` directory in the VIC repo and run `uncrustify`, which you can do by running the script in that directory, `./run_uncrustify.bash`. A simple `git diff --name-only` will list all files that `uncrustify` has updated the syntax for. If any files come up that you have edited, you should commit those syntax edits.

dgergel: Thanks @jhamman, just updated with your edits

PR #812, opened 2018-09-08
Fixed incorrect passing of lake_con structure to vic_run() and put_data().

tbohn:
- [x] closes #754 
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ran uncrustify prior to final commit
- [ ] ReleaseNotes entry

PR #814, opened 2018-09-08
Fix/less strict lake param validation

### LINKED ISSUE #813: Lake parameter validation should not be performed for cells in which lake_idx == -1
- Version: VIC 5.1.0
- Model settings: image driver; `LAKES == TRUE`
- Description:
When `LAKES==TRUE`, `vic_init()` attempts to validate the lake parameters of all grid cells, even if `lake_idx == -1` (which ostensibly should turn off lakes for that cell). As a result, lake parameters that are set to `nodata` values are treated as if they are intended to be real parameter values. Similarly, `vic_force()` looks for a `channel_in` forcing variable, but it is conceivable that the user intends for lakes to be fed only by local runoff within the grid cell and has not prepared a `channel_in` input variable. `channel_in` is not a standard meteorological forcing, so forcing the user to create a `channel_in` variable that is potentially 0 everywhere seems a little much.
- Proposed solution:
Make validation of lake parameters conditional on `lake_idx >= 0` and make reading of `channel_in` conditional on `param_set[CHANNEL_IN].SUPPLIED==TRUE`.

tbohn:
- [ ] closes #813
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry

NOTE: this PR is a branch off of PR #812. Please wait until PR #812 is resolved before evaluating this PR.

### DIFF
--- a/vic/drivers/image/src/vic_force.c
+++ b/vic/drivers/image/src/vic_force.c
@@ -186,14 +186,17 @@ vic_force(void)
     // Optional inputs
     if (options.LAKES) {
         // Channel inflow to lake
-        d3start[0] = global_param.forceskip[0] + global_param.forceoffset[0] +
-                     j;
-        get_scatter_nc_field_double(&(filenames.forcing[0]),
-                                    param_set.TYPE[CHANNEL_IN].varname,
-                                    d3start, d3count, dvar);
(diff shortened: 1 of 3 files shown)


tbohn:
This PR has been superceded by #826.

PR #817, opened 2018-09-11
Feature/add combined bulk density parameter option

### LINKED ISSUE #815: add parameter option for bulk density that combines organic and mineral content
Currently, the bulk density soil parameter represents the mineral soil bulk density. If the `ORGANIC_FRACT` option is set to `True`, then the model expects the parameters file to contain both mineral bulk density and the bulk density of the organic fraction of the soil. However, representative soil datasets rarely contain data specific to the organic fraction of soil, other than soil organic carbon content. For example, the ISRIC dataset has global bulk density data that combines mineral bulk density and the bulk density of the organic matter. 

We need a parameter option in VIC that allows for the combined (mineral + organic matter) bulk density to be directly read in as a parameter and then the organic matter bulk density can be derived from that and the mineral bulk density.

dgergel:
- [ ] closes #815
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry

This PR adds a global parameter file option `BULK_DENSITY_COMB` that enables combined bulk density (mineral and organic matter bulk density) to be read in from the parameters file.

### DIFF
--- a/vic/drivers/cesm/src/display_current_settings.c
+++ b/vic/drivers/cesm/src/display_current_settings.c
@@ -283,6 +283,12 @@ display_current_settings(int mode)
     else {
         fprintf(LOG_DEST, "ORGANIC_FRACT\t\tFALSE\n");
     }
+    if (options.BULK_DENSITY_COMB) {
+        fprintf(LOG_DEST, "BULK_DENSITY_COMB\t\tTRUE\n");
+    }   
+    else {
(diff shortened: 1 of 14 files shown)


bartnijssen:
@dgergel : Remember to make doc update (as a separate PR)

dgergel:
@bartnijssen just made the doc update in #824

PR #823, opened 2018-09-21
Fixes for fcanopy/LAI/esoil logic

### LINKED ISSUE #821: Esoil is 0 for pure bare soil tiles
- VIC version: 5.1.0
- Compiler and OS: n/a
- Problem: when VIC encounters pure bare soil tiles, and uses a default value of `fcanopy` of `1` (e.g., no `fcanopy` values were specified in the params file), no soil evap (`esoil`) occurs, and no ET occurs. The reason for this appears to be that, in `func_surf_energy_bal()`, the logic surrounding the call to `arno_evap()` only calls `arno_evap()` if `fcanopy < 1`. The intent of the logic was for tiles with sparse vegetation (`fcanopy < 1`) to have `esoil` take place over the fractional area `(1-fcanopy)`. However, this logic neglected the default case in which `fcanopy == 1` for all tiles.
- Proposed solution:
-- Remove the extra line in `vic_init()` that sets `fcanopy = 1` in all tiles; this line is unnecessary as it immediately follows an `if-else` block that sets `fcanopy` to either `1` for tile index < `Nveg - 1` or `MIN_FCANOPY` for index == `Nveg - 1`.
-- Additionally handle the case in which `LAI == 0`; in this case, we should set `fcanopy` to `0` as well so that `esoil` can take place.
-- Add logic around all places where we divide a variable by `fcanopy`, to not do the division if `fcanopy == 0`.
-- Add a missing multiplication by `fcanopy` for `Evap` and `Wdew` inside `func_canopy_energy_bal()`.

tbohn:
- [x] closes #821
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry

### DIFF
--- a/vic/vic_run/src/func_surf_energy_bal.c
+++ b/vic/vic_run/src/func_surf_energy_bal.c
@@ -750,7 +750,7 @@ func_surf_energy_bal(double  Ts,
     }
     Evap = 0.;
     if (!SNOWING) {
-        if (VEG && veg_var->fcanopy > 0) {
+        if (VEG) { // if VEG is true, then fcanopy > 0 and LAI > 0
             Evap = canopy_evap(layer, veg_var, true, veg_class, Wdew,
                                delta_t, NetBareRad, vpd, NetShortBare,
                                Tair, Ra_veg[1], elevation, rainfall,
--- a/vic/vic_run/src/func_surf_energy_bal.c
+++ b/vic/vic_run/src/func_surf_energy_bal.c
@@ -750,7 +750,8 @@ func_surf_energy_bal(double  Ts,
     }
     Evap = 0.;
     if (!SNOWING) {
-        if (VEG) { // if VEG is true, then fcanopy > 0 and LAI > 0
+        // if VEG is true, then fcanopy > 0 and LAI > 0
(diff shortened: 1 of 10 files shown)


dgergel:
@tbohn thanks for putting this together. I have tested this on a pure bare soil gridcell in the RASM domain and the behavior is what we want, e.g. this fixes the issue we were seeing before with no bare soil evaporation except during times of snow for pure bare soil gridcells. 

@bartnijssen and @jhamman, would either of you also like to do a review?

vic/vic_run/src/func_surf_energy_bal.c
```diff
@@ -750,7 +750,7 @@ func_surf_energy_bal(double  Ts,
     }
     Evap = 0.;
     if (!SNOWING) {
-        if (VEG && veg_var->fcanopy > 0) {
+        if (VEG) { // if VEG is true, then fcanopy > 0 and LAI > 0
```
dgergel: I would put comment above in a separate line for consistency

tbohn: Done.

Issue #412, opened 2016-03-22
ENH: Vegetation dependent snow albedo


jhamman:
As used in RASM Version 1.0 and detailed by Hamman et. al (In Revision).  See also RACM revision r370.

PR #835, opened 2018-10-01
Feature/add maximum snow albedo option

### LINKED ISSUE #834: add maximum snow albedo option
In VIC 4 versions, we were able to specify a maximum snow albedo that could be read in from the veg parameters file and used in place of the `param.SNOW_NEW_SNOW_ALB` option for all veg classes other than bare soil. We need to add this back in to VIC 5 so that it can be used for RASM domain simulations. 

#### Feature Requests

- Description of feature
- Description of the problem the feature addresses

See above. 

This feature will enable the Barlage et al 2005 dataset to be read in from the parameters file when the option `MAX_SNOW_ALBEDO` is set to True in the global parameter file
- Will the proposed feature be backward compatible? 

Yes, because it will be added in as an option. 
- Will the proposed feature change the science results of VIC?

When the option is set to true, yes - because this new snow albedo will be used across all snow routines for all veg types other than bare soil.

dgergel:
- [X] closes #834, see also #412 and #196 
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [x] ReleaseNotes entry

This PR adds in a vegetation-dependent maximum snow albedo for new snow option called `MAX_SNOW_ALBEDO`. When set to `True` in the global parameter file, maximum snow albedo is read in from the parameters file and used throughout the snow routines for new snow albedo except for bare soil. 

It also includes an unrelated bug fix to `vic/drivers/shared_image/src/vic_restore.c` which was necessary for being able to successfully compile the model.


tests/unit/vic_run/test_snow_utility.py
```diff
@@ -2,7 +2,8 @@
 from vic.vic import ffi
 from vic import lib as vic_lib
 
-
+# TODO: design new test to not only use the static parameter value
+@pytest.mark.xfail
```
jhamman: this should be fixed before we merge this.

dgergel:
@jhamman here is a point simulation for a gridcell that is in Siberia, with ~60% grassland (max snow albedo of 0.7) and ~35% bare soil: 

max_albedo_example

tests/unit/vic_run/test_snow_utility.py
bartnijssen: Agreed

vic/drivers/shared_image/src/vic_restore.c
```diff
@@ -970,7 +970,7 @@ check_init_state_file(void)
             dimlen = get_nc_dimension(&(filenames.init_state), "lake_node");
             if (dimlen != options.Nlakenode) {
                 log_err("Number of lake nodes in state file does not "
-                        "match LAKE_NODES (%d) in global parameter file",
+                        "match LAKE_NODES (%zu) in global parameter file",
```
bartnijssen: It's OK for now, but I would have preferred this as a separate PR to keep the individual PRs as focused as possible. Where this would break things is if we had to roll-back a PR for some reason. If we rolled back the `MAX_SNOW_ALBEDO` PR, that would also roll-back this unrelated bug fix. As said, leave this time.

dgergel:
@bartnijssen @jhamman I just took a stab at updating the unit test - not sure if this is the best way or not.

tests/unit/vic_run/test_snow_utility.py
```diff
@@ -2,11 +2,11 @@
 from vic.vic import ffi
 from vic import lib as vic_lib
 
-
+# updated to include veg-dependent albedo
 def test_snow_albedo_new_snow():
     assert vic_lib.snow_albedo(
         0.1, 1., 0.7, -77, 3600., 0,
-        ffi.cast('char', False)) == vic_lib.param.SNOW_NEW_SNOW_ALB
+        ffi.cast('char', False)) <= vic_lib.param.SNOW_NEW_SNOW_ALB
```
jhamman: I think you want to write a test that checks that the result of `snow_albedo` on timestep 0 is `max_snow_albedo`. You may also consider adding an additional test for when MAX_SNOW_ALBEDO is true or false.

dgergel: @bartnijssen and I just discussed this and I'm going to update this test in a future PR just so we can go ahead and merge the PR. I just created an issue for it; see #843 for reference.

jhamman: Okay. It sounds like you two have a plan in place. I'll just say that I think the best time/place to write tests is in the PR that changes specific bits of code. Its why we have a checkbox in the PR template for `new tests added`.

tests/unit/vic_run/test_snow_utility.py
bartnijssen: OK - we'll fix the test in future PR

PR #837, opened 2018-10-01
Feature/update organic fract option

### LINKED ISSUE #836: turn on organic fract option in VIC 5
In VIC 4, @tbohn added an `ORGANIC_FRACT` option in as part of his dissertation work, which, when set to true in the global parameter file, read in the bulk and soil density of organic matter as well as the fraction of organic matter in each layer of soil. This was then used in the frozen soils routine. We need to turn this on in VIC 5 to incorporate organic content of the soil into our RASM simulations. 

Additionally, we need to account for the new option `BULK_DENSITY_COMB` that enables us to read the combined bulk density (mineral and organic) in from the parameter file. If this is set to True, we want to specify that we read in the organic fraction and soil density of the organic matter, not the bulk density of the organic matter separately. 

- Description of feature
- Description of the problem the feature addresses

See above. 
- Will the proposed feature be backward compatible?

Yes. 
- Will the proposed feature change the science results of VIC?

If the `ORGANIC_FRACT` option is set to True, yes - as the organic fraction of the soil is used extensively in the frozen soils routine and will result in different soil temperatures in areas where the organic content of the soil is greater than 0.

dgergel:
- [X] closes #836
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [x] ReleaseNotes entry

This PR enables the `ORGANIC_FRACT` option to be turned on in VIC 5 (it was tested as part of this PR) and also enables the `BULK_DENSITY_COMB` option to be used in conjunction with it.


vic/drivers/shared_image/src/vic_init.c
```diff
@@ -588,16 +588,19 @@ vic_init(void)
...
+            // read in bulk_dens_org:
+            // organic bulk density for each soil layer
+            for (j = 0; j < options.Nlayer; j++) {
+                d3start[0] = j;
+                get_scatter_nc_field_double(&(filenames.params),
+                                            "bulk_density_org",
+                                            d3start, d3count, dvar);
+                for (i = 0; i < local_domain.ncells_active; i++) {
+                    soil_con[i].bulk_dens_org[j] = (double) dvar[i];
+                }
```
jhamman: what happens if `BULK_DENSITY_COMB == True`? Are you sure we're parsing `BULK_DENSITY_COMB` and `ORGANIC_FRACT` properly from the global parameter file?

dgergel: @jhamman yes. If `BULK_DENSITY_COMB == True`, then we read in the combined bulk density (mineral and organic) from the parameter file rather than reading in the organic bulk density separately. 

What I don't know is to what extent reading in the combined bulk density changes soil temperatures vis-a-vis turning on the `ORGANIC_FRACT` option, but I could test that and I think it would be useful to understand. Reading in the combined bulk density *without* turning on `ORGANIC_FRACT definitely impacts soil temperatures. As @bartnijssen and I have discussed, the bulk density is used throughout the frozen soils routine, so just using the combined bulk density without turning on `ORGANIC_FRACT` is already incorporating the organic content of the soil into the thermal conductivity calculations.

bartnijssen: Would be good to see a test. Can be done for a single grid cell.

dgergel:
I just made an issue for doing some follow-up analysis that will look at the impact on soil temperatures of turning on the `ORGANIC_FRACT` option. See #844 for reference

bartnijssen:
@dgergel : If you resolve the conflict then I can merge

dgergel:
@bartnijssen resolved the conflict earlier, this should be ready to merge

PR #854, opened 2018-12-01
Modify the type of seconds since midnight

quasi-mod:
- [ ] closes #xxx
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry

This PR changes the type of `seconds_since_midnight` and the explicit cast of `curr_date_time`. the previous implementation casted `curr_date_time`, which is a `time_t` value (64bit value), to `unsigned int` and took the mod. However, casting 64bit value to 32bit value (uint) will result in loosing upper 32 bits, and thus will result in the value wrapping around (btw, the exact year time_t value will wrap around in unsigned int is year 2106). 
To prevent this, `curr_date_time` should be casted to u64 type. 

Thanks in advance for your consideration.


jhamman:
Thank you for your contribution. This all looks fine to me. @tbohn / @yixinmao / @bartnijssen - care to take a peak?

PR #873, opened 2019-05-14
fixes for exact restarts and a few lines of cleanup

### LINKED ISSUE #872: fix exact restarts for CESM driver
Currently RASM testing of exact restarts with VIC5 shows that the following coupler fields need to be checked (as the global sums are different in exact restart tests): 
-- t
-- fv
-- taux
-- tauy
-- lat
-- sen
-- lwup
-- evap
-- rofliq

For additional reference, these variables have the same global sums in exact restart testing: 

-- albedo
-- tref
-- qref
-- logz0
--ram1
-- snowh
-- u10
-- swnet

dgergel:
- [ ] closes #872
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry

This PR is intended to fix our current problem with exact restarts in RASM with VIC5. For additional reference for details on which global sums are different, see #872. 

The solution proposed here is explicit typecasting to `double` as we currently do in `vic_store` for writing the model state. Here I have only explicitly typecasted the fields that are failing exact restarts, we may want to do this for all fields (??). 

This PR also includes two minor changes that I noticed while doing the typecasting - for latent and sensible heat we had kept the `+=` which was a relic from when we added over bands within `cesm_put_data` for these fields, this isn't needed anymore since we reconfigured this to use fields directly from `out_data`.

### DIFF
--- a/vic/drivers/cesm/src/cesm_put_data.c
+++ b/vic/drivers/cesm/src/cesm_put_data.c
@@ -118,7 +118,7 @@ vic_cesm_put_data()
         // that is already in the put_data routine
 
         // temperature, VIC: K, CESM: K
-        l2x_vic[i].l2x_Sl_t = out_data[i][OUT_RAD_TEMP][0];
+        l2x_vic[i].l2x_Sl_t = (double) out_data[i][OUT_RAD_TEMP][0];
 
         // albedo, VIC: fraction, CESM: fraction
         // Note: VIC does not partition its albedo, thus all types are
@@ -146,7 +146,7 @@ vic_cesm_put_data()
 
         // longwave up, VIC: W/m2, CESM: W/m2
         // adjust sign for CESM sign convention
-        l2x_vic[i].l2x_Fall_lwup = -1 *
+        l2x_vic[i].l2x_Fall_lwup = (double) -1 *
                                    (out_data[i][OUT_LWDOWN][0] -
                                     out_data[i][OUT_LWNET][0]);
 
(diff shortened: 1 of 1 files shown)


vic/drivers/cesm/src/cesm_put_data.c
```diff
@@ -146,27 +146,27 @@ vic_cesm_put_data()
...
         // turbulent heat fluxes
         // Note: both are the opposite sign from image driver
         // in accordance with the sign convention for coupled models
         // latent heat, VIC: W/m2, CESM: W/m2
-        l2x_vic[i].l2x_Fall_lat = -1 * out_data[i][OUT_LATENT][0];
+        l2x_vic[i].l2x_Fall_lat = (double) (-1 * out_data[i][OUT_LATENT][0]);
 
         // sensible heat, VIC: W/m2, CESM: W/m2
-        l2x_vic[i].l2x_Fall_sen += -1 * out_data[i][OUT_SENSIBLE][0];
+        l2x_vic[i].l2x_Fall_sen = (double) (-1 * out_data[i][OUT_SENSIBLE][0]);
```
bartnijssen: Why did you change the `+=`  to `=` here?

vic/drivers/cesm/src/cesm_put_data.c
```diff
@@ -146,27 +146,27 @@ vic_cesm_put_data()
...
 
         // evaporation, VIC: mm, CESM: kg m-2 s-1
-        l2x_vic[i].l2x_Fall_evap += -1 * out_data[i][OUT_EVAP][0] /
-                                    global_param.dt;
+        l2x_vic[i].l2x_Fall_evap = (double) (-1 * out_data[i][OUT_EVAP][0] /
```
bartnijssen: Why did you change the `+=` to `=` here?

vic/drivers/cesm/src/cesm_put_data.c
dgergel: This was leftover from when we were aggregating over bands - since we don't do that anymore, there's no reason to have the `+=`.

vic/drivers/cesm/src/cesm_put_data.c
dgergel: Same as above - this is leftover from when we were aggregating over bands.

dgergel:
@bartnijssen I don't think that the culprit is reading from/to a statefile, since we fixed exact restarts a while back in the image driver. I think what we're dealing with is purely within the CESM driver. 

I think properly initializing the `l2x` fields could be an issue. See `~/vic/drivers/cesm/src/vic_cesm_init_library.c` for how we are initializing the `l2x` fields with `SHR_CONST_SPVAL`. 

I also agree that specifying the literals is unlikely to solve the problem.

bartnijssen:
Agreed  on the `l2x` fields I will be looking at those. I am going to leave the PR open for the time being till I have had a chance to look in more detail at the initialization of the `l2x` fields. For the cesm driver we may need to add some additional fields to the state files for exact restarts.

Issue #890, opened 2019-08-20
Array indexing bug in support/VIC.4.2.glacier


anewman89:
Via slack discussion from @bartnijssen 

The problem happens in:
https://github.com/UW-Hydro/VIC/blob/support/VIC.4.2.glacier/src/glacier.c
in `int gl_volume_area(...)`. It turns out that the result of
https://github.com/UW-Hydro/VIC/blob/support/VIC.4.2.glacier/src/glacier.c#L89-L91
```
// time scaling
                ice_area_new = ice_area_old +
                               (ice_area_temp - ice_area_old) * exp(stepsize / BAHR_T);
```
CAN be a negative values for `ice_area_new` (I don’t know what that would physically signify.
If that is the case then `bot_band`, which gets set to `-1` in https://github.com/UW-Hydro/VIC/blob/support/VIC.4.2.glacier/src/glacier.c#L106, never gets updated and it results in a negative array index in https://github.com/UW-Hydro/VIC/blob/support/VIC.4.2.glacier/src/glacier.c#L119-L130 in either line 121 (if `GLACIER_OVERFLOW` is `TRUE`) or line 129 (if `GLACIER_OVERFLOW` is `FALSE`).
But negative array indices are not allowed I hear you protest - agreed, but since we do a lot of pointer arithmetic in C, I don’t think it is formally disallowed. The negative array index ends up clobbering something that should not be clobbered. Note that when I set `GLACIER_OUTFLOW` to `FALSE` and then replace line 129 with
`snow[iveg][0].gl_overflow = 0.; # instead of snow[iveg][bot_band].gl_overflow = 0.;`
then the model runs to completion on h2o (UW machine)
Note that this is not the solution. @jhamman: Should `ice_area_new` be allowed to be negative or how should that be handled to prevent mass balance errors? If we do not allow it to go negative, then what should the corresponding value of `bot_band` be?

From @jhamman:
I don’t think so. `if (ice_vol > 0.0)` (line 81) there should be a positive ice area.

`GLACIER_OVERFLOW` should only be non-zero (positive) when there are ice volume (area) exceeds the grid cell area (normalized). (edited) 

From @bartnijssen 
Agreed - note that `ice_area_temp = pow((ice_vol / BAHR_C), (1 / BAHR_LAMBDA));` (l.87) is not negative.
The negative value happens in the time scaling a few lines later:
```
/ time scaling
                ice_area_new = ice_area_old +
[...]

PR #891, opened 2019-08-20
Support/vic.4.2.glacier

anewman89:
This fixes issue #890.  The test case that was crashing on UW machine h2o, NCAR Cheyenne, and NCAR hydro-c1 now runs to completion on hydro-c1 with no negative `bot_band` values.

The fix simply ignores the time scaling when `ice_area_new` is negative and sets `ice_area_new = ice_area_old`.  Because this is within the positive `ice_vol` if statement (line 81), `ice_area_old is > 0` and the while statement on line 107 will be executed at least once moving `bot_band` to zero or larger.

This fix should probably still be tested on h2o and Cheyenne?  I can test on Cheyenne.

### DIFF
--- a/src/glacier.c
+++ b/src/glacier.c
@@ -90,6 +90,12 @@ gl_volume_area(snow_data_struct **snow,
                 ice_area_new = ice_area_old +
                                (ice_area_temp - ice_area_old) * exp(stepsize / BAHR_T);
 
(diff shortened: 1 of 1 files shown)


anewman89:
I rechecked for negative `bot_band` and found none in the test case on hydro-c1.

PR #894, opened 2019-09-24
cffi headers: respect $CC if defined

minrk:
default behavior is unchanged, but fixes builds on e.g. conda where compilers have funky names (https://github.com/conda-forge/vic-feedstock/pull/4) or `gcc` is not what should be used.

- [ ] closes #xxx
- [ ] tests passed
- [ ] new tests added
- [ ] science test figures
- [ ] ReleaseNotes entry

I ran uncrustify but didn't commit the changes because it changed a bunch of C files that aren't related to this PR. There isn't a release notes section for the next unreleased version, so I didn't feel ready to start one with such a minor fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,8 @@ matrix:
         - libnetcdf-dev
         - netcdf-bin
         - netcdf-doc
-        - gfortran
+        - gcc-4.8
+        - gfortran-4.8
         # open-mpi is built from source in cesm.travis
   # Classic Driver
   - compiler: clang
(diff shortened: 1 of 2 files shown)


jhamman:
Thanks @minrk - I'll take it from here. I'm really impressed by your persistence on this one. I'm guessing you aren't a user of the VIC model? Your efforts on conda-forge (and beyond) are much appreciated.

minrk:
>I'm guessing you aren't a user of the VIC model?

I'm not, this is with my conda-forge mpi maintainer hat on. I'm learning how to build and install lots of tools I don't use!

